PR #2, opened 2023-04-28
Added comments

pandey-tushar:


zxlive/rules.py
```diff
@@ -147,6 +151,12 @@ def Hadamard_slide(v):
 
 
 def add_node(u,v):
+    '''
+    g: BaseGraph[[VT,ET]]
+    u: node 1
+    v: node 2
+    returns: Adds a node between these two nodes and connect it via regular edge
```
y-richie-y, line 158: returns: A graph with a node between `u` and `v`, connected via regular edges

zxlive/rules.py
```diff
@@ -122,7 +122,11 @@ def bialgebra(v_list):
     return g
 
 def Hadamard_slide(v):
-    
+    '''
+    g: BaseGraph[[VT,ET]]
+    v: node or vertex 
+    returns: absorbs the hadamard appropriately if the node phase is pi/2
```
y-richie-y, line 128: Same here, the function returns a new graph with the `Hadamard absorbed if node phase is pi/2`

PR #9, opened 2023-05-07
Added qt-material to install_requires list

Econ01:
I'm not sure if this is correct way of adding the repositories therefore please make sure before confirming.


akissinger:
Yes, that's the correct way to add dependencies. I suggest adding a minimum versions while you are at it. The latest versions on PyPI are 2.14 for qt-material and 6.5 for PySide6.

PR #16, opened 2023-05-09
Clean rules

y-richie-y:
Simplify the rules a bit before implementing undo etc.

This also improves the correctness of the rules:
* `identity` was checking the wrong vertices
* `identity` should now work even when the edges have different types
* `fusion` used to also assume the other edges were simple edges and applying the hopf law

### DIFF
--- a/zxlive/rules.py
+++ b/zxlive/rules.py
@@ -3,6 +3,16 @@ from pyzx.utils import EdgeType, VertexType, toggle_edge, toggle_vertex
 import copy
 import numpy as np
 
+from pyzx import basicrules
+from pyzx import to_gh
+
+from statistics import fmean
+
+ET_SIM = EdgeType.SIMPLE
+ET_HAD = EdgeType.HADAMARD
+
+VT_Z = VertexType.Z
+VT_X = VertexType.X
 
 
 def phase_change(v, input):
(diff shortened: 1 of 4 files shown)


pandey-tushar:
For the fusion rule, when you raise exception warning, it's for the case when we can not have multi-edges? 

Otherwise, if I have a triangle with simple edge between v,w; simple edge between u,w and hadamard edge between u,v; then doing a v,w fusion should give us v,u with two edges, one h type and one simple.

y-richie-y:
I changed all the functions to use `basicrules.py` from `pyzx`.

`g.add_edge_smart` handles parallel edges correctly, we can have this until we support parallel edges.

zxlive/rules.py
pandey-tushar, line 1: In the color_change function, it does not go over a loop containing all the vertices selected? Or the function takes in only one argument now?

pandey-tushar:
Currently multi-edge is only supported for adding wires right? Not fusion or any other rule?

zxlive/rules.py
y-richie-y: I think in the case of color_change, it always only acted on one vertex. (Maybe I'm wrong) but in general I have made a few of the functions work on a single vertex or pair on vertices rather than a list. This will make the undo PR easier to do

y-richie-y:
I've replaced `add_edge` with `add_edge_smart` when appropriate.
I believe fusion also does the right thing as it uses `add_edge_smart` behind the scenes.

pandey-tushar:
1. Fusion looks good to me. For UNDO after fusion, it will undo the action or will it undo one fusion at a time? Because in the stack it's stored one by one.
2. Yes, the color change function is right. 
3. In the add_node function, if there are multiple edges, where will the node get added? Is it possible to select an edge?

y-richie-y:
1. It will undo everything at once at the moment.
3. At the moment there is no support for selecting an edge

pandey-tushar:
1. I see. The backup function adds the graph to the undo stack. DFS is definitely better than looking for every node pairs.
2. When multi-edges are allowed, then add_node function will/should not work ,correct? Or will/should it add a node to each edge?

y-richie-y:
Yeah, we will have to pick a specific edge rather than picking a pair of vertices.

pandey-tushar (approved):
Looks Good

Issue #4, opened 2023-05-02
Implementing reversible reset button.


Econ01:
Currently after user presses reset it is impossible to undo the reset. Adding this functionality is important.

akissinger:
I'm not sure "reset" will really be a meaningful thing to do once we implement loading and saving graphs.

y-richie-y:
This issue will be resolved by #18

Issue #11, opened 2023-05-09
Proper undo support


akissinger:
Adding proper undo support early is a good way to avoid a lot of pain/bugs later on. I suggest we replace the ad hoc implementation with a `QUndoStack`. Have a look at the Qt Undo Framework: https://doc.qt.io/qt-6/qundo.html

The normal way this is done is to create a new class extending [QUndoCommand](https://doc.qt.io/qt-6/qundocommand.html) for every user-initiated change to a graph. This class should override `redo` and `undo` for making the graph change and undo-in the graph change, respectively. Since you usually need some extra data to undo/redo a change, pass this in to the constructor and store it in the object. Note it is (usually) not necessary to keep a whole copy of the graph.

n.b. when you call `QUndoStack.push`, it automatically calls `redo` on the command being pushed, so the *only* place you should be changing the graph is in the `undo`/`redo` methods.

I know this seems a bit heavy duty, but it will save trouble in the long run. A couple of things to watch out for:
- pay attention to what's happening with the selection when undo-ing and redo-ing. Make sure the selection is changed/restored when things are removed or added back in and no non-existent things are ever selected
- make sure that the Undo/Redo commands correspond to currently visible graph, and not an old one or one on another tab (see suggestion to use tabs in #10)

PR #18, opened 2023-05-17
Use `QUndoCommand`s

y-richie-y:
Some more complicated transformations use `SetGraph` rather than transforming the graph minimally. This can be improved in future PRs.

### DIFF
--- a/zxlive/mainwindow.py
+++ b/zxlive/mainwindow.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 from PySide6.QtCore import QByteArray, QSettings
 from PySide6.QtGui import *
 from PySide6.QtWidgets import *
-from pyzx.utils import VertexType, toggle_vertex
+from pyzx.utils import VertexType
 import copy
 from fractions import Fraction
 
@@ -25,6 +25,7 @@ from fractions import Fraction
 from .graphview import GraphView
 from .rules import *
 from .construct import *
+from .commands import *
 
 from pyzx import basicrules
 from pyzx import to_gh
@@ -47,8 +48,6 @@ class MainWindow(QMainWindow):
         w.layout().setContentsMargins(0,0,0,0)
         w.layout().setSpacing(0)
         self.resize(1200, 800)
-        self.u_stack = []
-        self.d_stack = []
 
         # restore the shape/size of this window from the last time it was opened
         geom = conf.value("main_window_geometry")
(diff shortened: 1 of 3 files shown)


zxlive/mainwindow.py
```diff
@@ -66,16 +65,21 @@ def __init__(self) -> None:
         def Button_Fuse_Clicked():
             g, vs = self.get_elements()
             if vs == []:
+                self.graph_view.set_graph(g)
                 return
-            self.backup()
+
+            new_g = copy.deepcopy(g)
```
jvdwetering, line 71: Pyzx has a `BaseGraph.copy()` function. Do you maybe want to use that here instead? That copy function does change the labels of the vertices, so that might not be desirable.

y-richie-y:
Some edge additions are not reversible though, right? Would it always be possible to recover from the new graph? Or are you suggesting we compute the delta for g1 and g2, then g2 and g1?

jvdwetering:
What kind of edge additions are not reversible? Are you talking about things like a Hadamard self-loop resolving to a vertex having a pi phase added, and then to reverse this we would have to do two actions, get rid of the pi phase, and add the self-loop.

y-richie-y:
I haven't thought about it enough, but there seems to be a lot of cases with parallel edges. Are they all reversible?

akissinger:
I'm in favor of having lots of action-specific QUndoCommand's rather than implementing something generic. This is in line with the usual Qt GUI-writing style. If you have a piece of code that implements a GUI-generated change to your data structure, that code has to go somewhere, so you stick it in a new QUndoCommand with an inverting piece of code. The code gets run automatically when you push the command onto the undo stack.

This has the advantage that you are only changing the graph in one place (in the QUndoCommand), rather than changing e.g. in the event handler *and* in the undo/redo methods.

If you do decide that it's more elegant to have a general-purpose QUndoCommand for most things, I would just keep a whole copy of the graph. Deltas would only be relevant if memory was an issue, but these graphs are tiny. Just be careful to reset other bits of the GUI state (e.g. selected nodes) correctly when undo-ing and redo-ing.

y-richie-y:
Closes #4 and #11

zxlive/mainwindow.py
y-richie-y: Changing labels sounds troublesome, since the VItems carry the original labels of the vertices.

PR #24, opened 2023-05-26
Adhere to PEP 8 naming convention

### LINKED ISSUE #12: Consistent naming conventions
For consistency (and generally not getting weird-looking code), I suggest we adopt PEP 8 naming conventions where ever possible/reasonable:

https://peps.python.org/pep-0008/#naming-conventions

The main exception to this is the Qt bindings, which are all C++ style. The accepted convention is to use Python naming for your own code and Qt conventions just for calling/overriding Qt things.

In particular, capital letters and underscores should (almost) never go together. This looks very strange to me.

Also, I suggest avoiding writing the type of a thing in variable names, e.g. the many variables called `Button_....`. The exception to this is if there is only one thing of a given type around, e.g. a `graph_view` of type `GraphView`.

mark-koch:
This closes #12

### DIFF
--- a/zxlive/mainwindow.py
+++ b/zxlive/mainwindow.py
@@ -63,7 +63,7 @@ class MainWindow(QMainWindow):
         self.graph_view.set_graph(construct_circuit())
         # self.graph_view.set_graph(zx.generate.cliffords(5, 5))
 
-        def Button_Fuse_Clicked():
+        def fuse_Cclicked():
             g, vs = self.get_elements()
             if vs == []:
                 self.graph_view.set_graph(g)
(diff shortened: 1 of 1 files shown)


zxlive/mainwindow.py
```diff
@@ -63,7 +63,7 @@ def __init__(self) -> None:
         self.graph_view.set_graph(construct_circuit())
         # self.graph_view.set_graph(zx.generate.cliffords(5, 5))
 
-        def Button_Fuse_Clicked():
+        def fuse_Cclicked():
```
jvdwetering, line 66: Is this a typo?

jvdwetering (commented):
It looks like you only applied this naming convention to this one file? Are the other files already fine?

zxlive/mainwindow.py
mark-koch: Yeah good catch 👍

jvdwetering:
Just checking, but have you made sure that the program still works and none of the buttons crash the program?
This is where unit tests would really have come in handy of course.

PR #30, opened 2023-06-04
WIP: drag node command

y-richie-y:
I've turned dragging into a command. Since dragging is continuous, using `undo_stack.begin_macro` and `undo_stack.end_macro` groups this into a single command.

Problem:
The `MoveNode` `QUndoCommand` takes `v, old_pos, new_pos` in the constructor, which means this is incompatible with the `SetGraph` command.

Any ideas?

### DIFF
--- a/zxlive/commands.py
+++ b/zxlive/commands.py
@@ -171,3 +171,19 @@ class ChangeColor(QUndoCommand):
         self.graph_view.set_graph(g)
 
     undo = redo = toggle
+
+
+class MoveNode(QUndoCommand):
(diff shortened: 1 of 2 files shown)


mark-koch:
Why not send the command once the dragging is done (i.e. in `mouseReleaseEvent`)? I already implemented this in #31 (saw your PR too late, sorry)

y-richie-y:
The macro solves the problem of multiple commands, this issue is more to do with how the command isn't associated with a graph scene but a specific node.

PR #35, opened 2023-06-06
Refactor diagram import and export code

mark-koch:
I noticed that the diagram import/export code from #25 contains a lot of repeated code. This PR does some refactoring:
* Move import/export/error dialogs to a seperate file `dialogs.py`
* Define enum `FileFormat` for supported file format instead of relying on stringly typing
* Utility method `show_error_msg` to display error messages

### DIFF
--- a/zxlive/edit_panel.py
+++ b/zxlive/edit_panel.py
@@ -1,15 +1,13 @@
 from fractions import Fraction
 from typing import Iterator
 
-from PySide6.QtCore import QFile, QIODevice, QTextStream
-from PySide6.QtWidgets import QToolButton, QFileDialog, QMessageBox, QInputDialog
+from PySide6.QtWidgets import QToolButton, QInputDialog
 from pyzx import EdgeType, VertexType
 from pyzx.graph.base import BaseGraph, VT
 
-import pyzx as zx
-
 from .base_panel import BasePanel, ToolbarSection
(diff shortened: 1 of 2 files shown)


zxlive/edit_panel.py
```diff
@@ -79,20 +77,15 @@ def _vert_double_clicked(self, v: VT) -> None:
...
-            new_phase = Fraction(input)
+            new_phase = Fraction(input_)
         except ValueError:
-            msg = QMessageBox()
-            msg.setIcon(QMessageBox.Critical)
-            msg.setText("Wrong Input Type")
-            info_text = "Please enter a valid input (e.g. 1/2, 2)"
-            msg.setInformativeText(info_text)
-            msg.exec_()
+            show_error_msg("Wrong Input Type", "Please enter a valid input (e.g. 1/2, 2)")
```
mark-koch, line 88: I noticed that this diaglog disappears immediately (even before I made any changes). Can any of your reproduce this?

After some investigating it seems like the `QMessageBox` breaks if it is put after `QInputDialog.getText(...)`

jvdwetering:
Yeah it is probably best to split it up into other files like this.
I think it would also be good to have a "any supported filetype" option, but that is an additional feature while this is a refactor, so that is maybe best reserved for another PR.

PR #36, opened 2023-06-06
Small panel refactors

mark-koch:
Some small improvements following on from #31
* Better `ToolbarSection` constructor
* Turn commands into dataclasses

### DIFF
--- a/zxlive/base_panel.py
+++ b/zxlive/base_panel.py
@@ -20,6 +20,10 @@ class ToolbarSection:
     buttons: Sequence[QToolButton]
     exclusive: bool = False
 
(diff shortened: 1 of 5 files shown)


jvdwetering:
What does changing these commands into dataclasses do?

mark-koch:
Dataclasses give us `__eq__` for free. Normally, they also generate `__init__` but there was an issue with the `QUndoCommand` base class.

But I figured out how to make it work (see latest commit), so now we don't have to write constructors for commands

jvdwetering:
Okay, is this fine for merging now?

mark-koch:
Yes, should be good to go 👍

PR #37, opened 2023-06-07
Improved selection

mark-koch:
* Use Qt builtin methods for spider selection and movement
* Add selection using rectangular drag
* Support moving all selected spiders at once


jvdwetering:
By the way, I think I can just add you to the project @mark-koch, so that you don't have to make PRs all the time, or so that you can at least approve them yourself.

mark-koch:
That would be nice, thanks!

PR #51, opened 2023-07-19
force graph to be up to date in edge if it changes type

DNA386:
very ugly feeling way to fix edge type update not showing (but also I'm not sure where else the graph should be getting updated, since it seems like it may need to be fully recreated to get the update (eg when spider fusing it))


mark-koch:
I'd much rather store the `GraphScene` in the `EItem` similar to what `VItem` is doing (see #56)

PR #54, opened 2023-07-23
Proof step viewer

mark-koch:
This PR adds a model for proofs and a proof step viewer for the rewrite mode

### DIFF
--- a/zxlive/commands.py
+++ b/zxlive/commands.py
@@ -1,15 +1,21 @@
 from dataclasses import dataclass, field
 from fractions import Fraction
-from typing import Optional, Iterable, Union
+from typing import Optional, Iterable, Union, List, Any
 import copy
 
+from PySide6.QtCore import QItemSelection, QModelIndex, QItemSelectionModel, \
+    QSignalBlocker
 from PySide6.QtGui import QUndoCommand
+from PySide6.QtWidgets import QListView
 
 from pyzx import basicrules
+from pyzx.graph import GraphDiff
 from pyzx.utils import EdgeType, VertexType
 
 from .common import VT, ET, GraphT
 from .graphview import GraphView
+from .proof import ProofModel, Rewrite
+
 
 @dataclass
 class BaseCommand(QUndoCommand):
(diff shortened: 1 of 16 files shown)


zxlive/commands.py
```diff
@@ -238,3 +244,92 @@ def toggle(self) -> None:
...
+        # Remove steps from the proof model until we're at the currently selected step
+        self._old_selected = self.step_view.currentIndex().row()
+        self._old_steps = []
+        for _ in range(self.proof_model.rowCount() - self._old_selected - 1):
+            self._old_steps.append(self.proof_model.pop_rewrite())
+
+        diff = self.diff or GraphDiff(self.g, self.new_g)
+        self.proof_model.add_rewrite(Rewrite(self.name, diff), self.new_g)
+
+        # Select the added vertex
```
jvdwetering, line 279: "vertex" -> "step"?

zxlive/commands.py
```diff
@@ -238,3 +244,92 @@ def toggle(self) -> None:
...
+
+
+@dataclass
+class GoToRewriteStep(SetGraph):
+    """Shows the graph at some step in the proof.
+
+    Undoing returns to the previously selected proof step.
+    """
+
+    def __init__(self, graph_view: GraphView, step_view: QListView, old_step: int, step: int) -> None:
```
jvdwetering, line 310: I thought there was some problem with redefining the __init__ in a dataclass?

zxlive/proof.py
```diff
@@ -0,0 +1,99 @@
...
+        """Adds a rewrite step to the model."""
+        self.beginInsertRows(QModelIndex(), len(self.graphs), len(self.graphs))
+        self.graphs.append(new_graph)
+        self.steps.append(rewrite)
+        self.endInsertRows()
+
+    def pop_rewrite(self) -> tuple[Rewrite, GraphT]:
+        """Removes the latest rewrite from the model.
+
+        Returns the rewrite and the grap that previously resulted from this rewrite.
```
jvdwetering, line 89: grap -> graph

zxlive/proof_panel.py
```diff
@@ -114,7 +153,7 @@ def _remove_id(self, v: VT) -> None:
         new_g = copy.deepcopy(self.graph)
         basicrules.remove_id(new_g, v)
         anim = anims.remove_id(self.graph_scene.vertex_map[v])
-        cmd = SetGraph(self.graph_view, new_g)
+        cmd = AddRewriteStep(self.graph_view, new_g, self.step_view, "id")
```
jvdwetering, line 156: "remove identity"

zxlive/proof_panel.py
```diff
@@ -128,5 +167,87 @@ def _unfuse(self, v: VT, left_neighbours: list[VT]) -> None:
...
+    circle_outline_width = 3
+
+    def paint(self, painter: QPainter, option: QStyleOptionViewItem, index: Union[QModelIndex, QPersistentModelIndex]) -> None:
+        painter.save()
+
+        # Draw background
+        painter.setPen(Qt.GlobalColor.transparent)
+        if option.state & QStyle.State_Selected:
+            painter.setBrush(QColor(204, 232, 255))
+        # TODO: Why is hover not working?
```
jvdwetering, line 201: It looks like it is working fine for me?

jvdwetering (commented):
Very nice!
I gave it a bit of a test run, and I noticed that if you redo twice, that it breaks. It doesn't give an exception, so not sure what is going wrong in the logic.

But in any case: this is really cool.

jvdwetering:
A thing where I don't know what we exactly want to do: if you go back to a previous step and move some vertices, and then switch to a different step, it is not remembered that you moved those vertices. I'm not sure if that would be desired behaviour though.
Something which I think might be nice (though maybe not for a first version), is that you can decide to move a vertex at an earlier step, and then propogate this move forwards to all the future steps. Kind of like a 'refactor'. This would be useful if you realise that you actually need more space between your spiders in later diagrams. This has always been a pain to do in tikzit.

zxlive/commands.py
mark-koch: We'll run into issues once we try to subclass `GoToRewriteStep` but this works fine for now

zxlive/proof_panel.py
mark-koch: I think this might be a MacOS thing, but I found a fix. We have to call `setAttribute(Qt.WA_Hover)` on the viewport of the list view. Now it also works on my machine 👍

mark-koch:
> I gave it a bit of a test run, and I noticed that if you redo twice, that it breaks. It doesn't give an exception, so not sure what is going wrong in the logic.

This should (hopfully) be fixed now

> if you go back to a previous step and move some vertices, and then switch to a different step, it is not remembered that you moved those vertices. I'm not sure if that would be desired behaviour though.

For now I implemented that the position is saved only for the current step. But I agree, it would be nice if the position were  propagated to future steps (though a way to opt out of this would probably be needed as well). Anyways, I'd leave that to another PR.

I think this one is now ready to merge if you're happy with it?

PR #74, opened 2023-08-08
Custom rewrites

RazinShaikh:
This PR contains preliminary implementation of custom rewrites. Currently, the rewrites only work with exact phases but we can add support for more general types of phases in the future.

### DIFF
--- a/zxlive/edit_panel.py
+++ b/zxlive/edit_panel.py
@@ -166,6 +166,14 @@ class GraphEditPanel(BasePanel):
 
     def _vert_double_clicked(self, v: VT) -> None:
         if self.graph.type(v) == VertexType.BOUNDARY:
+            input_, ok = QInputDialog.getText(
+                self, "Input Dialog", "Enter Boundary Index:"
+            )
+            try:
+                input_ = int(input_.strip())
+                self.graph.set_vdata(v, "boundary_index", input_)
+            except ValueError:
+                show_error_msg("Wrong Input Type", "Please enter a valid input (e.g. 1, 2)")
             return
 
         input_, ok = QInputDialog.getText(
--- a/zxlive/edit_panel.py
+++ b/zxlive/edit_panel.py
@@ -167,11 +167,11 @@ class GraphEditPanel(BasePanel):
     def _vert_double_clicked(self, v: VT) -> None:
         if self.graph.type(v) == VertexType.BOUNDARY:
             input_, ok = QInputDialog.getText(
-                self, "Input Dialog", "Enter Boundary Index:"
+                self, "Input Dialog", "Enter Qubit Index:"
             )
             try:
                 input_ = int(input_.strip())
-                self.graph.set_vdata(v, "boundary_index", input_)
+                self.graph.set_qubit(v, input_)
             except ValueError:
                 show_error_msg("Wrong Input Type", "Please enter a valid input (e.g. 1, 2)")
             return
--- a/zxlive/mainwindow.py
+++ b/zxlive/mainwindow.py
@@ -229,7 +229,7 @@ class MainWindow(QMainWindow):
             self.close()
         if not self.active_panel.undo_stack.isClean():
             name = self.tab_widget.tabText(i).replace("*","")
-            answer = QMessageBox.question(self, "Save Changes", 
+            answer = QMessageBox.question(self, "Save Changes",
                             f"Do you wish to save your changes to {name} before closing?",
                             QMessageBox.StandardButton.Yes | QMessageBox.StandardButton.No | QMessageBox.StandardButton.Cancel)
             if answer == QMessageBox.StandardButton.Cancel: return False
--- a/zxlive/mainwindow.py
+++ b/zxlive/mainwindow.py
@@ -20,8 +20,10 @@ import copy
 
 from PySide6.QtCore import QFile, QFileInfo, QTextStream, QIODevice, QSettings, QByteArray, QEvent
 from PySide6.QtGui import QAction, QShortcut, QKeySequence, QCloseEvent
-from PySide6.QtWidgets import QMessageBox, QMainWindow, QWidget, QVBoxLayout, QHBoxLayout, QTabWidget, QFileDialog, QSizePolicy
+from PySide6.QtWidgets import QMessageBox, QMainWindow, QWidget, QVBoxLayout, QTabWidget, QDialog, QFormLayout, QLineEdit, QTextEdit, QPushButton, QDialogButtonBox
 from pyzx.graph.base import BaseGraph
+
+from zxlive import proof_actions
 
 from .commands import AddRewriteStep
 
@@ -83,13 +85,13 @@ class MainWindow(QMainWindow):
         close_action = self._new_action("Close", self.close_action, QKeySequence.StandardKey.Close,
             "Closes the window")
         close_action.setShortcuts([QKeySequence(QKeySequence.StandardKey.Close), QKeySequence("Ctrl+W")])
-        # TODO: We should remember if we have saved the diagram before, 
+        # TODO: We should remember if we have saved the diagram before,
         # and give an open to overwrite this file with a Save action
         save_file = self._new_action("&Save", self.save_file, QKeySequence.StandardKey.Save,
             "Save the diagram by overwriting the previous loaded file.")
         save_as = self._new_action("Save &as...", self.save_as, QKeySequence.StandardKey.SaveAs,
             "Opens a file-picker dialog to save the diagram in a chosen file format")
-        
+
         file_menu = menu.addMenu("&File")
         file_menu.addAction(new_graph)
         file_menu.addAction(open_file)
@@ -143,6 +145,10 @@ class MainWindow(QMainWindow):
         view_menu.addAction(zoom_out)
         view_menu.addAction(fit_view)
 
+        create_new_rewrite = self._new_action("Create new rewrite", self.create_new_rewrite, None, "Create a new rewrite")
+        rewrite_menu = menu.addMenu("&Rewrite")
+        rewrite_menu.addAction(create_new_rewrite)
+
         simplify_actions = []
         for simp in simplifications.values():
             simplify_actions.append(self._new_action(simp["text"], self.apply_pyzx_reduction(simp), None, simp["tool_tip"]))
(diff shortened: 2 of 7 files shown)


zxlive/mainwindow.py
```diff
@@ -350,7 +356,47 @@ def reduce() -> None:
...
+                return
+            rewrite = proof_actions.ProofAction.from_dict({
+                "text":name.text(),
+                "tooltip":description.toPlainText(),
+                "matcher": proof_actions.create_custom_matcher(self.left_graph),
+                "rule": proof_actions.create_custom_rule(self.left_graph, self.right_graph),
+                "type": proof_actions.MATCHES_VERTICES,
+            })
+            proof_actions.rewrites.append(rewrite)
+            dialog.accept()
```
jvdwetering, line 396: Qt does weird stuff with closures in a callback. Check that this continues to work when you add multiple rules, and that dialog doesn't suddenly refer to the previously constructed dialog and gives some kind of segfault.

zxlive/proof_actions.py
```diff
@@ -112,17 +115,90 @@ def rewriter() -> None:
...
             self.btn_group.addButton(btn)
             action.button = btn
 
     def update_active(self, g: GraphT, verts: List[VT], edges: List[ET]) -> None:
         for action in self.actions:
-            action.update_active(g,verts,edges)
+            action.update_active(g, verts, edges)
+
+
+def to_networkx(graph: Graph) -> nx.Graph:
```
jvdwetering, line 131: Using networkx is for now I think fine, but when we want to do matching where we also have variables and parameters, then we might need to roll our own implementation.

zxlive/proof_actions.py
```diff
@@ -112,17 +115,90 @@ def rewriter() -> None:
...
+def custom_matcher(graph: Graph, in_selection: Callable[[VT], bool], lhs_graph: nx.Graph) -> List[VT]:
+    verts = [v for v in graph.vertices() if in_selection(v)]
+    subgraph_nx, _ = create_subgraph(graph, verts)
+    graph_matcher = GraphMatcher(lhs_graph, subgraph_nx,\
+        node_match=categorical_node_match(['type', 'phase'], default=[1, 0]))
+    if graph_matcher.is_isomorphic():
+        return verts
+    return []
+
+def custom_rule(graph: Graph, vertices: List[VT], lhs_graph: nx.Graph, rhs_graph: nx.Graph) -> pyzx.rules.RewriteOutputType[ET,VT]:
```
jvdwetering, line 165: So if I understand the logic of this function, you remove all the vertices in the original graph, and then you replace them with new ones (except for the boundaries). While this does give you correct semantics, I think in terms of positioning information we will want to use something that tries to match the lhs and the rhs as match as possible, and only adds/removes those vertices that really have to be removed.
tl;dr: The matching graphs we are using have position information, and this right now is not being used.

jvdwetering: This might not be the thing to do for the first draft, but some kind of positioning must be done, because you dont want them all to appear just on top of each other.

zxlive/mainwindow.py
```diff
@@ -350,7 +356,47 @@ def reduce() -> None:
             cmd = AddRewriteStep(self.active_panel.graph_view, new_graph, self.active_panel.step_view, reduction["text"])
             self.active_panel.undo_stack.push(cmd)
         return reduce
-    
+
+    def create_new_rewrite(self) -> None:
```
jvdwetering, line 360: It might be nicer to move most of this logic to its own file, like how the other messageboxs are in the dialogs.py file (so maybe could just put it there for now?)

jvdwetering (commented):
Looks good! I have some small comments, and there's definitely room for enhancements in the future.
Maybe something to consider as well is to do a check for the semantics of the lhs and rhs when you build the new rule. Since we are not considering bang-boxes or any fancy logic we just have to calculate the matrix of the lhs and of the rhs and see if they are equal, and we can warn the user if they are trying to create a rewrite rule that doesn't preserve semantics. 
We could even use this to automatically find the right scalar factor to scale the graph by when you do the graph replacement.

zxlive/edit_panel.py
```diff
@@ -166,6 +166,14 @@ def _vert_moved(self, vs: list[tuple[VT, float, float]]) -> None:
 
     def _vert_double_clicked(self, v: VT) -> None:
         if self.graph.type(v) == VertexType.BOUNDARY:
+            input_, ok = QInputDialog.getText(
+                self, "Input Dialog", "Enter Boundary Index:"
+            )
+            try:
+                input_ = int(input_.strip())
+                self.graph.set_vdata(v, "boundary_index", input_)
```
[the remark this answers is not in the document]
RazinShaikh: That makes sense. How should we get this ordered input/output information from the user in the UI?

zxlive/mainwindow.py
RazinShaikh: I checked with multiple rules. For some reason, the previous rewrite stops working after adding a new rewrite. The matcher works fine but the rule does not. I'll need to investigate but I am not sure if it is related to Qt.

zxlive/proof_actions.py
RazinShaikh: The networkx implementation for matching graphs looks quite involved. It would be better to not reimplement it if it is not necessary. The networkx graph matcher takes in a semantic_feasibility function as an optional argument. It is worth checking if passing a semantic_feasibility function is enough to deal with variables and parameters.

zxlive/proof_actions.py
RazinShaikh: I was wondering about this. Where is this position information? I don't see anything obvious in pyzx Graph data structure.

zxlive/mainwindow.py
jvdwetering: I'm pretty sure that is Qt's doing. It doesn't like it if you use a lambda as a callback. If you replace it by named functions, for instance by def'ing a function inside another function and then return'ing that function, that it might fix it. You can see a couple of examples of this in mainwindow.py I believe.

zxlive/edit_panel.py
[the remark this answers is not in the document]
jvdwetering: So there is an g.auto_detect_io() function that automatically tries to match the order of the boundary vertices to and input output list, so that could be a starting point. But then we need some kind of logic for when the user tries to change this. I'm not sure what the best way would be to do this.

zxlive/mainwindow.py
RazinShaikh: This was just some silly bug where we need to call `ProofActionGroup.copy()`. Fixed in https://github.com/Quantomatic/zxlive/pull/74/commits/516d9f08344374cdc19dfa449d002c7de3e41254

zxlive/proof_actions.py
jvdwetering: you can call g.get_pos(v) I believe, or otherwise g.get/set_qubit(v), which corresponds to the y coordinate and g.get/set_row(v) which corresponds to the x coordinate.

zxlive/edit_panel.py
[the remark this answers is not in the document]
RazinShaikh: How reliable is g.auto_detect_io()? Could it lead to situations where it detects io differently on left and right sides of the rewrite?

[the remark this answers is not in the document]
jvdwetering: It does the following logic: for every boundary, check wether the thing it is connected to is to its left or to its right. If it is to its right, then it is an input, otherwise it is an output. I think it raises an exception if it is pointed vertically. Then for the ordering, it simply goes top to bottom, based on the g.get_qubit(v) property.

zxlive/proof_actions.py
RazinShaikh: That did not work for me. I got the attribute error: `'GraphS' object has no attribute 'get_pos'`. Same for `get_qubit` and `get_row`.

zxlive/edit_panel.py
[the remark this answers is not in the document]
RazinShaikh: That seems reasonable but it does force the user into the left to right convention, which is fine for now. Anyway, let's get rid of boundary_index and use the input/output tuple instead.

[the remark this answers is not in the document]
RazinShaikh: This was actually quite straightforward. Implemented in https://github.com/Quantomatic/zxlive/pull/74/commits/eb7929f0302c18443d4b4c71b69539a143cb2b55

zxlive/proof_actions.py
akissinger: It could be that the networkx implementation is involved because it is doing a clever graph iso algorithm. Basic iso checking via depth-first traversal is not hard to implement, and would probably be fast enough.

RazinShaikh: You're right, it implements the VF2++ algorithm (https://doi.org/10.1016/j.dam.2018.02.018). We can do our own implementation when we start looking at variables and parameters, but let's do that after the [poly](https://github.com/Quantomatic/zxlive/tree/poly) branch is merged.

zxlive/proof_actions.py
```diff
@@ -197,6 +206,19 @@ def custom_rule(graph: Graph, vertices: List[VT], lhs_graph: nx.Graph, rhs_graph
 
     return etab, vertices_to_remove, [], True
 
+def get_vertex_positions(graph, rhs_graph, boundary_vertex_map):
+    pos_dict = {v: (graph.qubit(m), graph.row(m)) for v, m in boundary_vertex_map.values()}
```
jvdwetering, line 210: So one thing I want to point out, is that .qubit() is a y coordinate and .row() is an x coordinate. So this mapping you have is (y,x) which might lead to confusion later. It's fine here though as you using it consistently.
Otherwise this looks like a very neat solution.

RazinShaikh: Good point, let me change that.

RazinShaikh: Fixed in https://github.com/Quantomatic/zxlive/pull/74/commits/495749992f5e5fbc37431990c51d6ffccbc83c3d

RazinShaikh:
@jvdwetering I believe I have addressed all of your comments now. Let me know if there is anything else to do for this PR.

jvdwetering:
Yes this all looks good, and in any case we can still add stuff to it later of course.

PR #77, opened 2023-08-12
W node

RazinShaikh:


zxlive/commands.py
```diff
@@ -12,6 +12,8 @@
 from pyzx.graph import GraphDiff
 from pyzx.utils import EdgeType, VertexType
 
+from zxlive.w_node import get_w_partner, is_w_edge_type, is_w_vertex_type
```
jvdwetering, line 15: Shouldn't this be `from .w_node import ...`?

jvdwetering:
I don't really see where you are changing the VertexType and EdgeType classes to allow for W-like things. Is this a change you are making in a local copy of pyzx?

RazinShaikh:
It's in my fork here: https://github.com/RazinShaikh/pyzx/tree/non-flexsymmetric

zxlive/commands.py
RazinShaikh: both works. this was probably auto-imported by VS-code. I will change it to `from .w_node` for consistency.

zxlive/commands.py
```diff
@@ -89,17 +89,42 @@ class ChangeNodeColor(BaseCommand):
...
+        for v1 in self.vs:
+            if vertex_is_w(self.g.type(v1)):
+                v2 = get_w_partner(self.g, v1)
+                v2_neighbors = [v for v in self.g.neighbors(v2) if v != v1]
+                for v3 in v2_neighbors:
+                    self.g.add_edge(self.g.edge(v1,v3), edgetype=self.g.edge_type(self.g.edge(v2,v3)))
+                self._old_w_info[v1] = {'partner': v2,
+                                        'partner_type': self.g.type(v2),
+                                        'partner_pos': (self.g.row(v2), self.g.qubit(v2)),
+                                        'neighbors': v2_neighbors}
```
mark-koch, line 125: Consider using a `NamedTuple` instead of a dictionary here. This would require less "stringly-typing"

zxlive/commands.py
```diff
@@ -89,17 +89,42 @@ class ChangeNodeColor(BaseCommand):
...
+                self.g.remove_vertex(v2)
+            self.g.set_type(v1, self.vty)
```
mark-koch, line 127: Now that this type change can also be a W node, we should rename the command from `ChangeNodeColor` to something like `ChangeNodeType`

zxlive/commands.py
```diff
@@ -120,6 +145,8 @@ def undo(self) -> None:
     def redo(self) -> None:
         self._old_etys = [self.g.edge_type(e) for e in self.es]
         for e in self.es:
+            if self.g.edge_type(e) == EdgeType.W_IO:
+                continue
```
mark-koch, line 149: Note sure if this check belongs here. I would expect a command that changes edge colors to actually affect all specified edges. Maybe we should ignore IO edges before invoking the command?

zxlive/commands.py
```diff
@@ -142,6 +169,27 @@ def redo(self) -> None:
...
+
+    def undo(self) -> None:
+        assert self._added_input_vert is not None
+        assert self._added_output_vert is not None
+        self.g.remove_vertex(self._added_input_vert)
+        self.g.remove_vertex(self._added_output_vert)
+        self.update_graph_view()
+
+    def redo(self) -> None:
+        self._added_input_vert = self.g.add_vertex(VertexType.W_INPUT, self.y-0.3, self.x)
```
mark-koch, line 189: I don't like the hardcoded 0.3 here. Maybe make it a global constant or default member of the dataclass (i.e. declare `input_offset: float = 0.3`)?

zxlive/edit_panel.py
```diff
@@ -98,6 +100,8 @@ def create_icon(self, shape: str, color: str) -> QIcon:
             painter.drawEllipse(4, 4, 56, 56)
         elif shape == "square":
             painter.drawRect(4, 4, 56, 56)
+        elif shape == "triangle":
```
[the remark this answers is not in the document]
mark-koch: And the same thing for `VERTICES` and `EDGES` dict keys

zxlive/edit_panel.py
```diff
@@ -153,10 +157,19 @@ def _ety_clicked(self, ety: EdgeType.Type) -> None:
...
+        if self._curr_vty == VertexType.W_OUTPUT:
+            cmd = AddNonFlexNode(self.graph_view, x, y)
+        else:
+            cmd = AddNode(self.graph_view, x, y, self._curr_vty)
         self.undo_stack.push(cmd)
 
     def _add_edge(self, u: VT, v: VT) -> None:
+        if vertex_is_w(self.graph_scene.g.type(u)) and vertex_is_w(self.graph_scene.g.type(v)):
+            if self.graph_scene.g.edge_type(self.graph_scene.g.edge(u, v)) == EdgeType.W_IO:
+                return
```
mark-koch, line 169: I'm not sure I understand what this check is doing here?

zxlive/edit_panel.py
```diff
@@ -200,20 +215,29 @@ def paste_graph(self, graph: GraphT) -> None:
...
+            if self.graph_scene.g.edge_type((v1, v2)) == EdgeType.W_IO:
+                rem_vertices.extend([v1, v2])
+        if not rem_vertices and not selected_edges: return
         new_g = copy.deepcopy(self.graph_scene.g)
         self.graph_scene.clearSelection()
         new_g.remove_edges(selected_edges)
-        new_g.remove_vertices(selection)
-        cmd = SetGraph(self.graph_view,new_g) if len(selection) > 128 \
+        new_g.remove_vertices(list(set(rem_vertices)))
+        cmd = SetGraph(self.graph_view,new_g) if len(rem_vertices) > 128 \
```
mark-koch, line 232: nit: I guess you should check the length of the set here

zxlive/vitem.py
```diff
@@ -364,3 +415,20 @@ def refresh(self) -> None:
...
+        math.cos(angle) * (p.x() - origin.x()) - math.sin(angle) * (p.y() - origin.y()) + origin.x(),
+        math.sin(angle) * (p.x() - origin.x()) + math.cos(angle) * (p.y() - origin.y()) + origin.y()
+    )
+
+def get_w_partner_vitem(g: GraphT, graph_scene: GraphScene, v: VT) -> VItem:
+    """Get the VItem of the partner of a w_in or w_out vertex."""
+    assert g.type(v) in {VertexType.W_INPUT, VertexType.W_OUTPUT}
+    partner = get_w_partner(g, v)
+    if partner not in graph_scene.vertex_map:
+        return None
```
mark-koch, line 433: So the return type should be `Optional[VItem]`?

zxlive/vitem.py
```diff
@@ -256,6 +301,12 @@ def mouseReleaseEvent(self, e: QGraphicsSceneMouseEvent) -> None:
             return
         if e.button() == Qt.MouseButton.LeftButton:
             if self._old_pos != self.pos():
+                if self.g.type(self.v) == VertexType.W_INPUT:
+                    # set the position of w_in to next to w_out at the same angle
+                    w_out = get_w_partner_vitem(self.g, self.graph_scene, self.v)
+                    w_in_pos = w_out.pos() + QPointF(0, 0.3 *SCALE)
```
mark-koch, line 307: Since the `0.3` also shows up here (and not only in `AddNonFlexNode`) we should definitely make it a constant

zxlive/edit_panel.py
```diff
@@ -200,20 +215,29 @@ def paste_graph(self, graph: GraphT) -> None:
...
-        if not selection and not selected_edges: return
+        rem_vertices = selection.copy()
+        for v in selection:
+            if vertex_is_w(self.graph_scene.g.type(v)):
+                for n in self.graph_scene.g.neighbors(v):
+                    if self.graph_scene.g.edge_type((v, n)) == EdgeType.W_IO:
+                        rem_vertices.append(n)
+        for v1, v2 in selected_edges:
+            if self.graph_scene.g.edge_type((v1, v2)) == EdgeType.W_IO:
+                rem_vertices.extend([v1, v2])
```
mark-koch, line 226: Maybe we should make it that `W_IO` edges cannot be selected? This could be done easily by calling
```python
self.setFlag(QGraphicsItem.GraphicsItemFlag.ItemIsSelectable, self.g.edge_type(self.e) != EdgeType.W_IO)
```
in `EItem.refresh()`

zxlive/vitem.py
```diff
@@ -256,6 +301,12 @@ def mouseReleaseEvent(self, e: QGraphicsSceneMouseEvent) -> None:
...
+                    w_in_pos = rotate_point(w_in_pos, w_out.pos(), w_out.rotation())
+                    self.setPos(w_in_pos)
```
mark-koch, line 309: Don't we also have to update the position in the graph?

zxlive/commands.py
RazinShaikh: The W_IO edge is not a real edge, as in it does not have any semantics so it never makes sense to change a W_IO edge to any other type. But I am happy to move this check to _ety_clicked method instead.

RazinShaikh: Moving it to _ety_clicked is actually better because it skips adding empty command to the undo stack. Now the ChangeEdgeType command does not have this check: 843bdb667890bc59c1470b9ded6c12e538176133

zxlive/edit_panel.py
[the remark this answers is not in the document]
RazinShaikh: Resolved in 7d88c118e1e16d8eca3abb17b8742b1dfd7de954

zxlive/commands.py
```diff
@@ -142,6 +169,27 @@ def redo(self) -> None:
         self._added_vert = self.g.add_vertex(self.vty, self.y, self.x)
         self.update_graph_view()
 
+@dataclass
+class AddNonFlexNode(BaseCommand):
+    """Adds a new non-flexsymmetric node at a given position."""
```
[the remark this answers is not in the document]
RazinShaikh: For now, let's make it `AddWNode` because I don't have any other non flexsymmetric node in mind to add anytime soon.

zxlive/edit_panel.py
RazinShaikh: I think this is a great idea but why should it be in `EItem.refresh()` instead of `__init__()`?

zxlive/edit_panel.py
RazinShaikh: That is indeed hard to parse. Here's a cleaner version of the function:
```python
def _add_edge(self, u: VT, v: VT) -> None:
    g = self.graph_scene.g
    if vertex_is_w(g.type(u)) and get_w_partner(g, u) == v:
        return
    if g.type(u) == VertexType.W_INPUT and len(g.neighbors(u)) >= 2 or \
       g.type(v) == VertexType.W_INPUT and len(g.neighbors(v)) >= 2:
        return
    cmd = AddEdge(self.graph_view, u, v, self._curr_ety)
    self.undo_stack.push(cmd)
```
Let me know if this makes sense.

zxlive/vitem.py
RazinShaikh: This is done inderectly in the lines 315 - 318, where the vertices_moved signal is emitted
```python
else:
    scene.vertices_moved.emit([
        (it.v, *pos_from_view(it.pos().x(),it.pos().y()))
        for it in scene.selectedItems() if isinstance(it, VItem)
    ])
```

RazinShaikh: This code actually needed some modification to also update the w partner in the graph. Fixed in https://github.com/Quantomatic/zxlive/pull/77/commits/8dd6533acdd12690c9c67a04ed9c62d3ad238334

zxlive/edit_panel.py
RazinShaikh: I have added it in `__init__` and it seems to work

RazinShaikh:
@mark-koch Thanks a lot for your comments. I have gone through each of them and incorporated your suggestions. Let me know if it looks good to you.

zxlive/edit_panel.py
mark-koch: When the edge type changes for some reason, selection should be reanabled. Not sure if IO edges can change (maybe a rewrite or though the user), but we could put it into `refresh` just to be safe

zxlive/edit_panel.py
mark-koch: Way clearer now, thank you!

zxlive/vitem.py
mark-koch: That makes sense 👍

zxlive/edit_panel.py
RazinShaikh: They should never change but happy to move it to `refresh` to be safe.

jvdwetering:
Also, when you add new W-spiders, depending on where you click the orientation of the W-spider changes, sometimes in weird ways: 
image

jvdwetering:
I don't understand the bug I outlined above, I tried with two different configurations:
image
And the left one rewrites correctly with fuse W nodes, but the right one doesn't:
image

RazinShaikh:
Hi John, that is indeed not the right behaviour. For instance, if you drag the nodes, they will fuse correctly. I will check later today and see what's going wrong when you press the fuse button.

RazinShaikh:
@jvdwetering the fusion bug should be fixed by this PR https://github.com/Quantomatic/pyzx/pull/129 in PyZX.

RazinShaikh:
> Also, when you add new W-spiders, depending on where you click the orientation of the W-spider changes, sometimes in weird ways: 
image

Fixed in https://github.com/Quantomatic/zxlive/pull/77/commits/306754b3263e23e4160308a505d677be3cde6cdd. The fix for this is to call `set_vitem_rotation()` when a graph is updated. It should always have the correct orientation now.

RazinShaikh:
@mark-koch @jvdwetering are you guys happy for this to be merged?

jvdwetering:
It looks like you got a new conflict because Boldi fixed some mypy bugs, but once that's fixed I think this can be merged in.

PR #89, opened 2023-08-26
Add requirements.txt.

dlyongemallo:


dlyongemallo:
Some of the version numbers can probably be adjusted down, if there's some reason to do so.

RazinShaikh:
Looks good

PR #93, opened 2023-08-30
Add buttons to toolbar for undo/redo. Fixes #73.

### LINKED ISSUE #73: Add buttons in toolbar for undo/redo
When you are working with a tablet, it's not very nice to have to go to the Edit menu to click undo/redo, and it might be hard to use the keyboard to ctrl+Z. I suggest adding buttons in the toolbar (for instance next to the select/vertex/edge buttons in edit mode, and next to the select/magic wand tools in Proof mode). These should also have nice icons.

dlyongemallo:
The icons were drawn by myself using inkscape.

### DIFF
--- a/zxlive/base_panel.py
+++ b/zxlive/base_panel.py
@@ -1,14 +1,13 @@
 from dataclasses import dataclass
 from typing import Iterator, Sequence, Optional
 
-from PySide6.QtCore import QSize
-from PySide6.QtGui import QIcon
+from PySide6.QtGui import QAction
 from PySide6.QtWidgets import QWidget, QVBoxLayout, QToolBar, QToolButton, QButtonGroup, \
     QSplitter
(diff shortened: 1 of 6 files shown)


jvdwetering:
Thanks! This does introduce a bit of duplication of code though, since there is now logic dealing with the undo/redo command in mainwindow and in base_panel. I think it is possible to link the buttons you create to the QAction that is defined in mainwindow. Then all the info about undo/redo is in mainwindow. Although you might still need to have the logic to disable/enable the buttons in base_panel itself.
See e.g.: https://stackoverflow.com/questions/9342868/qaction-vs-qtoolbutton-and-when-to-override-the-basic-class

dlyongemallo:
I've unified the logic between the toolbar buttons and the menu items, but to do so I've had to pass in the undo/redo `QAction`s to the panels.

Issue #71, opened 2023-08-07
UI improvement: Rewrite rules


RazinShaikh:
The side panel can have the rewrites, simplification routines and the proof steps. The code for sidebar of edit mode might be useful for implementing this.

RazinShaikh:
Using tabs sounds like a nice idea :)

boldar99:
We could do something like that of VSCode.

Rewrites on the side with collapsable groups. We could also implement searching and scrolling.

![image](https://github.com/Quantomatic/zxlive/assets/49209650/49858b00-7d68-40ba-8aa7-ca7d0b020bd1)

RazinShaikh:
We should also allow the user to delete custom rewrites from the UI.

jvdwetering:
I've now added the tabbed interface for different groups of rewrites.
![image](https://github.com/Quantomatic/zxlive/assets/15610543/1126c427-f9bb-464f-b173-e145b1db0165)

jvdwetering:
I think we could also add different tab names based on subfolders in the folder containing the custom rewrite rules. This would make sense if we expect people to be adding a lot of custom rewrite rules.

A design decision we want to consider is whether a rewrite rule can appear in multiple tabs. For instance, should Z-spider fusion only appear in Basic rules, or also in the ZXW rule set and the ZH rule set?

RazinShaikh:
Is it possible to add an option to change from horizontal tabs to vertical tabs? Most browsers support this. That would look a bit like Boldi's suggestion.

RazinShaikh:
@jvdwetering can you also add these simplifications as a tab?

![image](https://github.com/Quantomatic/zxlive/assets/28257517/807ab9d8-6cb3-42d2-8aa8-0c9701e59ebf)

jvdwetering:
> Is it possible to add an option to change from horizontal tabs to vertical tabs? Most browsers support this. That would look a bit like Boldi's suggestion. I fell that horizontal list of rewrites is a bit inconvenient if you have a lot of custom rewrites because then you need to scroll horizontally.

I believe there is some built in Qt feature that allows you to switch being a horizontal layout and a vertical layout.

RazinShaikh:
I have added these simplification routines in this commit https://github.com/Quantomatic/zxlive/commit/a3d6260512b342577ca458029e18c5c6a1d03c1c

RazinShaikh:
[...]

PR #94, opened 2023-08-30
Rule editor UI

### LINKED ISSUE #83: UI for creating custom rewrites
ZXLive supports adding custom rewrites. To do so, we can choose a graph file for the left and the right-hand side of the equality. It would be a better experience to have a window for creating custom rewrites. This window should be split vertically, where the user can draw the left and the right-hand side of the rewrite.

RazinShaikh:
Closes #83

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build
 *.json
 *.zxp
 .env
+*.zxr
--- a/zxlive/base_panel.py
+++ b/zxlive/base_panel.py
@@ -42,10 +42,10 @@ class BasePanel(QWidget):
     file_path: Optional[str]
     file_type: Optional[FileFormat]
 
-    def __init__(self, graph: GraphT, graph_scene: GraphScene, *actions: QAction) -> None:
+    def __init__(self, *actions: QAction) -> None:
         super().__init__()
-        self.graph_scene = graph_scene
-        self.graph_view = GraphView(self.graph_scene)
+        self.graph_scene = None
+        self.graph_view = None
         self.actions = actions
         self.undo_stack = AnimatedUndoStack(self)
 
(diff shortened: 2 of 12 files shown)


jvdwetering:
Looks good! Something that's a little confusing: When you have the rule editor open, you can still save a graph. But I think what this does is just save the currently selected graph. I would have expected this to save the whole rule as a lemma to a file. Currently all the rules are stored in a single file, but if we want users to be able to share lemmas between each other, we probably want each rule to be in a separate file. So maybe instead of having`customrules.json`, there could be a designated folder which holds all the lemmas?
So maybe we want some logic to determine what the type of thing is that we are saving, a graph or a rule.
Anyway, this is all not so important for this PR.

jvdwetering:
If you click on 'Save rule' while the diagram is not well-formed, for instance if it has a boundary node not connected to anything, then it throws an exception instead of telling the user the diagram is not well-formed.
It also allows you to save rules which are not well-typed, that have for instance a different amount of boundaries on the LHS then on the RHS. If you do this and match the LHS, it simply doesn't match which is weird behaviour. It probably would be best to just pop up a message saying that the rule is not well-typed.

jvdwetering:
Also maybe something not for this PR: Have a button that allows the user to switch the LHS and RHS. Or maybe something that allows you to automatically create the reverse rule?

RazinShaikh:
> Also maybe something not for this PR: Have a button that allows the user to switch the LHS and RHS. Or maybe something that allows you to automatically create the reverse rule?

@boldar99 is working on the side panel for rewrite rules (#71). We can have a button there to apply the rewrite rule in the forward or backward direction.

RazinShaikh:
@jvdwetering the rules are now saved to individual files and the save button in the file menu has the right behavior. I have also added a function to check the rule and warn the user.

PR #100, opened 2023-08-31
Update shape when changing vertex type. Fixes #57.

### LINKED ISSUE #57: Changing a spider to H-box does not update shape
Similar to issue #49, when drawing a diagram and changing a spider to an Hbox, the color is updated but the shape remains round.
This is probably because the vitem path needs to be changed to a rectangle rather than an ellipse.

dlyongemallo:
There's some weird behaviour with W-nodes if `W_INPUT` rather than `W_OUTPUT` is the selected node when its type is changed, and it's also not possible to change another node type into a W-node. This behaviour already exists, but the change introduced in this PR makes it more visually obvious.

### DIFF
--- a/zxlive/vitem.py
+++ b/zxlive/vitem.py
@@ -104,27 +104,6 @@ class VItem(QGraphicsPathItem):
         self.setFlag(QGraphicsItem.GraphicsItemFlag.ItemIsSelectable, True)
         self.setFlag(QGraphicsItem.GraphicsItemFlag.ItemSendsGeometryChanges, True)
 
-        pen = QPen()
-        pen.setWidthF(3)
-        pen.setColor(QColor(BLACK))
-        self.setPen(pen)
-
-        path = QPainterPath()
-        if self.ty == VertexType.H_BOX:
-            path.addRect(-0.2 * SCALE, -0.2 * SCALE, 0.4 * SCALE, 0.4 * SCALE)
-        elif self.ty == VertexType.W_OUTPUT:
-            #draw a triangle
-            path.moveTo(0, 0.2 * SCALE)
-            path.lineTo(0.25 * SCALE, -0.15 * SCALE)
-            path.lineTo(-0.25 * SCALE, -0.15 * SCALE)
-            path.lineTo(0, 0.2 * SCALE)
-        elif self.ty == VertexType.W_INPUT:
-            scale = 0.3 * SCALE
-            path.addEllipse(-0.2 * scale, -0.2 * scale, 0.4 * scale, 0.4 * scale)
-        else:
-            path.addEllipse(-0.2 * SCALE, -0.2 * SCALE, 0.4 * SCALE, 0.4 * SCALE)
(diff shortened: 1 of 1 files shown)


RazinShaikh:
Can you describe the weird behavior when `W_INPUT` is selected? I found that when both `W_INPUT` and `W_OUTPUT` are selected, it throws an error but when only `W_INPUT` is selected, it works fine.

RazinShaikh:
This happens because changing the node type of `W_INPUT` or `W_OUTPUT` also removes its w partner. When both `W_INPUT` and `W_OUTPUT` are selected, it acts on them twice. A fix is to simply remove the w partner from the list of selected vertices.
I tried the following change in `ChangeNodeType` class in commands.py and it seems to work.
```diff
def redo(self) -> None:
    if self._old_w_info is None:
        self._old_w_info = {}
    if vertex_is_w(self.vty):
        return
+   self.vs = list(self.vs)
+   for v in self.vs:
+       if vertex_is_w(self.g.type(v)):
+           v2 = get_w_partner(self.g, v)
+           if v2 in self.vs:
+               self.vs.remove(v2)
    self._old_vtys = [self.g.type(v) for v in self.vs]
    for v1 in self.vs:
        if vertex_is_w(self.g.type(v1)):
            v2 = get_w_partner(self.g, v1)
            v2_neighbors = [v for v in self.g.neighbors(v2) if v != v1]
            for v3 in v2_neighbors:
                self.g.add_edge(self.g.edge(v1,v3), edgetype=self.g.edge_type(self.g.edge(v2,v3)))
            self._old_w_info[v1] = self.WInfo(partner=v2,
                                              partner_type=self.g.type(v2),
                                              partner_pos=(self.g.row(v2), self.g.qubit(v2)),
                                              neighbors=v2_neighbors)
            self.g.remove_vertex(v2)
        self.g.set_type(v1, self.vty)
    self.update_graph_view()
```

dlyongemallo:
Thanks. I think if a W-node is selected (either the input or output or both), changing the vertex type should always act on the input vertex (triangle) rather than the output (little circle). It's where its "centre of mass" is. 

Also, changing a vertex to W-node seemingly does the change without a visual update. Selecting a node, then clicking the "W node" button, seems to do nothing, but the vertex change to W is on the undo stack, and if I try to `undo` it, the assertion in the first line (`assert self._old_vtys is not None`) fails. It seems that the W-node isn't completely implemented yet. What's missing to make clicking on "W node" also visually update the vertex?

jvdwetering:
Probably related to this: would it make sense to just not make the W_INPUT selectable? Or do we want to keep it selectable so that you can rotate the W in the direction you want it to face?

RazinShaikh:
@jvdwetering We want to keep it selectable to rotate the W.
@dlyongemallo Thanks for pointing out the issues with changing the type of W-node. Since this PR does not change any node type changing behavior, we should merge it and open a separate issue for W node bugs.

PR #96, opened 2023-08-31
Deploy

RazinShaikh:
@boldar99 shall we merge this?

PR #106, opened 2023-09-01
Fix `NameError` by adding back import of `List` from `typing` to `commands.py`.

dlyongemallo:


dlyongemallo:
Would it be possible to enable continuous integration so this type of breakage doesn't happen again?

RazinShaikh:
Yes, that would be great. @boldar99 was trying to set it up in the deploy branch (see #96) but he hasn't had time to finish it yet.

PR #111, opened 2023-09-21
Proper bug fix for indexing error in `construct_circuit`.

dlyongemallo:
`nvlist` is created with the vertex type element set using index `i+qubits-1`, where `i` is not in scope (and has the value 3 from the previous loop). This bug was previously fixed by ignoring this element in `nvlist` and adding a separate list `ty` to keep track of the vertex type, and explicitly using the `VertexType.{Z,X}` constants, but this isn't necessary if the correct index is used.

Also fix a mypy warning and sort `elist`.

### DIFF
--- a/zxlive/construct.py
+++ b/zxlive/construct.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Tuple
 
 from pyzx.utils import EdgeType, VertexType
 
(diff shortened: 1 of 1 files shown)


RazinShaikh:
Thanks, merging this. We will remove construct.py at some point and start the application with an empty canvas.

Issue #121, opened 2023-09-28
vertex 0 is never animated


dlyongemallo:
Repro:

1. perform an operation that involves vertex number 0

Expected result: an animation is shown

Actual result: an animation is not shown

RazinShaikh:
This should be fixed now that the PRs are merged. Shall we close it?

dlyongemallo:
Yes. Thanks!

PR #122, opened 2023-09-28
Fix a bug where vertex 0 is never animated.

dlyongemallo:
Closes issue #121.


dlyongemallo:
Ah, I messed up the fix because the original code had an assignment expression, and I had originally split it into assignment then if, but combined them like the original code when I pushed the commit. But the assignment and if really do need to be two separate lines for a proper fix.

RazinShaikh:
Does merging https://github.com/Quantomatic/zxlive/pull/123 fix this?

dlyongemallo:
Yes, either #123 or #124 (or both). The assignment expression is fine for python 3.8 and higher, and as discussed zxlive is supporting >= 3.9 so it's okay. (I had removed it in an earlier state of this code because I wasn't sure whether 3.7 was going to be supported or not.)

PR #129, opened 2023-10-01
Temporarily disable `is_well_formed` check.

dlyongemallo:
This should be re-enabled after https://github.com/Quantomatic/pyzx/pull/162 is merged.


RazinShaikh:
I was working with a local copy of pyzx and enabled the well-formed check in zxlive too soon.

Issue #117, opened 2023-09-27
some actions should be disabled when last tab is closed


dlyongemallo:
Repro steps:

1. close the last tab
2. open the File menu, and select Close, Save, or Save as

Expected result: nothing happens (or, even better, those items are disabled and not possible to select)

Actual: assertion failure `assert self.active_panel is not None`

Also, the assert is unnecessary for `close_action` as `self.active_panel` is never referenced there. Indeed, the code after the assert explicitly handles the case where there are no tabs, in which case it closes the app. The assert that's now there prevents the app from being closed.

dlyongemallo:
The same comment applies to the Edit and View menu actions as well.

PR #130, opened 2023-10-01
Disable Save and Save as when last tab is closed.

dlyongemallo:
Also allow Close to close the app when there are no tabs.

Parts of issue #117.


RazinShaikh:
This does not work for me. I am able to click save after the last tab is closed and it just throws an assertion error.

dlyongemallo:
Thanks for catching that. It should work now when clicking on the close button with the mouse.

RazinShaikh:
nice, that works!

PR #131, opened 2023-10-01
Fix a bunch of mypy errors.

dlyongemallo:
Also use python 3.9 style type hinting consistently for generic types (list, tuple, dict).

### DIFF
--- a/zxlive/animations.py
+++ b/zxlive/animations.py
@@ -1,6 +1,6 @@
 import itertools
 import random
-from typing import Optional, Callable, Tuple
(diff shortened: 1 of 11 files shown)


dlyongemallo:
The remaining mypy errors count should go from 44 down to 10 after this PR. The only remaining errors are a bit more tricky: 8 are due to the forward declaration of `BaseGraph` in `pyzx`, and 2 are functions returning the results of functions from other libraries for which mypy doesn't know the type.

boldar99 (approved):
Looks good to me!

PR #138, opened 2023-10-02
Simplify code to set up new tab action.

dlyongemallo:
Reading the code, I'm pretty sure the intention with `new_tab` was just to add a new shortcut key, and since a mechanism was added to do that in #118, might as well use it.

I think the string `"Reinitialize with an empty graph"` must've been a leftover from when the app didn't have tabs (and so the `new_graph` action would re-initialize the only window).


boldar99 (approved):
LGTM

PR #140, opened 2023-10-03
Disable certain Edit and View menu items when last tab is closed.

### LINKED ISSUE #117: some actions should be disabled when last tab is closed

dlyongemallo:
Re-enable when new tab is created.

(Choosing Edit or View menu items after the last tab has been closed previously resulted in assertion failures.)

Fixes #117.

### DIFF
--- a/zxlive/mainwindow.py
+++ b/zxlive/mainwindow.py
@@ -106,18 +106,16 @@ class MainWindow(QMainWindow):
         file_menu.addAction(self.close_action)
         file_menu.addAction(self.save_file)
         file_menu.addAction(self.save_as)
-        self.save_file.setEnabled(False)
-        self.save_as.setEnabled(False)
 
         self.undo_action = self._new_action("Undo", self.undo, QKeySequence.StandardKey.Undo,
             "Undoes the last action", "undo.svg")
         self.redo_action = self._new_action("Redo", self.redo, QKeySequence.StandardKey.Redo,
             "Redoes the last action", "redo.svg")
-        self.cut_action = self._new_action("Cut", self.cut_graph,QKeySequence.StandardKey.Cut,
+        self.cut_action = self._new_action("Cut", self.cut_graph, QKeySequence.StandardKey.Cut,
             "Cut the selected part of the diagram")
-        self.copy_action = self._new_action("&Copy", self.copy_graph,QKeySequence.StandardKey.Copy,
+        self.copy_action = self._new_action("&Copy", self.copy_graph, QKeySequence.StandardKey.Copy,
             "Copy the selected part of the diagram")
-        self.paste_action = self._new_action("Paste", self.paste_graph,QKeySequence.StandardKey.Paste,
+        self.paste_action = self._new_action("Paste", self.paste_graph, QKeySequence.StandardKey.Paste,
             "Paste the copied part of the diagram")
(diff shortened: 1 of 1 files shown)


zxlive/mainwindow.py
```diff
@@ -173,6 +168,29 @@ def __init__(self) -> None:
...
+        self.deselect_all_action.setEnabled(has_active_tab)
+        self.zoom_in_action.setEnabled(has_active_tab)
+        self.zoom_out_action.setEnabled(has_active_tab)
+        self.fit_view_action.setEnabled(has_active_tab)
+        self.show_matrix_action.setEnabled(has_active_tab)
+
+        # Paste is enabled only if there is something in the clipboard.
+        self.paste_action.setEnabled(has_active_tab and self.copied_graph is not None)
+
+        # Undo and redo are always disabled whether on a new tab or closing the last tab.
```
RazinShaikh, line 187: The undo/redo buttons should be enabled or disabled based on what is in the undo stack. It feels a bit weird to hard code them to be disabled.

dlyongemallo: I believe this is the correct behaviour. The undo/redo stack is tied to the tab (and should be reset to disabled when switching tabs), unlike paste which is tied to the clipboard (and whose state is therefore preserved when switching tabs).

On closing the last tab, previously only Save and Save As were disabled, but undo/redo should be disabled as well when there is no tab, which is what this PR changes. So this PR fixes a bug where previously the undo/redo could be enabled even when there's nothing to undo/redo.

On `new_graph` (whether opening the app for the first time, creating a new graph, or opening a graph from file), this PR preserves the existing behaviour. In `_new_panel`, where calls were previously made to disable undo/redo, a call to `_reset_menus` is made instead. This call is made just prior to the `canUndoChanged` and `canRedoChanged` connections are made, meaning that it's acting on the state of the undo stack before the new panel is created. Once the panel is created, a call to `tab_changed` will enable/disable undo/redo based on the undo stack. On a new tab, the undo stack is in any case clear, so these will be disabled. If in the future, the undo stack is saved with a file for example, then on opening a file, after `_reset_menus` is called, the call to `tab_changed` will in any case update the undo/redo menu items to the correct state. 

So what this PR does is set the undo/redo menu items to a predictable state when initialising, which is then overridden when an actual change to the undo stack (new graph, file open, switch tabs, various Edit commands) updates them later. This is good defensive programming, because if a bug is later introduced that breaks the undo stack, we can be sure that every time we create a new graph or open a file to try to repro the bug, the undo/redo buttons start in a disabled state until they are explicitly set with a user action that updates them.

(If these buttons are made to depend on the undo stack here, they will be acting on the previous state of the undo stack before `_new_panel` is finished. So I think that would actually be the incorrect behaviour.)

RazinShaikh: Thanks for the explanation :)

Issue #148, opened 2023-10-19
`string_to_phase` is fragile and breaks on some inputs


dlyongemallo:
When taking in an input which should result in a `Fraction`, the `string_to_phase` function is a bit inconsistent in what inputs it will accept, and also raises an exception (instead of catching it and showing a dialogue) on certain inputs. On inputs ending in `*pi`, it will silently return the wrong result (due to stripping `pi` and leaving `*` as part of the expression).

jvdwetering:
Can this be closed now?

dlyongemallo:
Two major things noted above are fixed: when the input is accepted, the result will be correct, and if it is not accepted, there is an error message (and not silent failure or crash). 

There is still a minor inconsistency because of the special handling of `*pi`: while both `3pi` and `3*pi` are accepted, `3*alpha` is accepted while `3alpha` is not, despite the fact that when the input is `3*alpha`, the phase is actually _displayed_ as `3alpha` on the diagram itself. Thus, intuitively, a user might expect to be able to _enter_ the input as `3alpha`. 

I can follow up with a PR to fix that, and we can leave this open until then.

dlyongemallo:
While looking at the grammar for poly, I noticed two things. 

While a fractional phase can be expressed as a float, e.g., `3/2` and `1.5` are both acceptable (likewise `3/2*pi` and `1.5*pi`), this isn't allowed when there's a variable, i.e., `3/2*alpha` is okay, but `1.5*alpha` is not. I'm not sure if that's an oversight or if it's by design that the factor has to be an integer or exact fraction when variables are involved.

Also, I noticed that `Poly` can express an exponent, e.g., `a^2`, but there's nothing in the parser grammar to handle that. Is that intended to be implemented? That is, will a phase ever be something like `a^2`?

jvdwetering:
My feeling is that the allowed possibilities for variables should be restrictive, as we need special behaviour for the parameters. However, I think allowing fractional multiples of variables should generally be fine. Powers of variables are a different question. Generally, if the variable is a Boolean, the power can be ignored since it does not do anything. If it is not a Boolean, then the exponent does matter, but I don't know of any nice way we could do rewriting with them. So perhaps the best thing is to just disallow exponents in the expression altogether.

dlyongemallo:
Okay, then I think this issue can be closed now. The phase input dialog now accepts all reasonable input, and only really malformed inputs cause an error now (and which all result in an error popup to the user).

PR #149, opened 2023-10-19
Make `string_to_phase` slighty more robust.

dlyongemallo:
Add a test for various input formats which are expected to result in a `Fraction`.

See issue #148.

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 pytest
-mypy zxlive
+mypy zxlive test
--- a/test/test_editor_base_panel.py
+++ b/test/test_editor_base_panel.py
@@ -23,7 +23,7 @@ from zxlive.poly import Poly, Term, Var, new_var
 
 
 def test_string_to_phase():
-    types_dict = {'a': False, 'b': False}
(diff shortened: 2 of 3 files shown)


boldar99:
We could add a couple of tests for the complex case. The rest seems good!

dlyongemallo:
I've added a basic test for the complex input case, but there's something I'm not sure about. When the `fraction_or_complex` argument has the value `complex` and `str` contains variable parameters, e.g., `a + bj`, what the intended result? Is the `Poly` case with variable parameters intended to be used only for a `fraction` input (and thus an error should be shown if a `complex` input is given with variable parameters), or is the intention to treat `j` as the imaginary unit in the `complex` case?

jvdwetering:
I don't think we have really thought about variables in the complex case. So I think for now not allowing users to do this would be the way to go.

dlyongemallo:
Okay, I've refactored the code to disallow variables for complex inputs. I've also split the fraction and complex input logic into separate functions so it's clear what's allowed for each.

jvdwetering:
Is this ready to be merged?

dlyongemallo:
Yes, ready to be merged.

Issue #156, opened 2023-10-30
dialog windows are sometimes hidden behind main window


dlyongemallo:
In `dialogs.py`, the various dialogs don't always take a `parent` parameter consistently. In particular, `show_error_msg` doesn't take a `parent` at all, and error messages can sometimes show up behind the main window which is rendered inactive, which is a confusing state for the user.

Also: in other dialog-showing functions, the `parent` parameter is sometimes typed `QWidget` and sometimes typed `MainWindow`, and it's sometimes the first parameter and sometimes the last. The API would be nicer if this was made consistent.

valleyofblackpanther:
Hi, @dlyongemallo @RazinShaikh check this commit [commit](https://github.com/valleyofblackpanther/zxlive/commit/62b8f5f8c0882dddbf8ef9cedc8b5925d8abdc39)
1. I have modified the `show_error_msg` function to accept a parent parameter. This parameter is then used to specify the parent of the `QMessageBox` so that the error message box is modal to the parent. 
2. In each function that invokes `show_error_msg`, i ensured to pass the parent parameter. This means I added a parent argument to the signature of those fucntions if it's not already there. I did modification in the function `import_diagram_from_file`.

RazinShaikh:
The original problem of dialog windows being hidden behind the main window is now solved. I have renamed the issue as per @dlyongemallo 's suggestion. The following still needs to be updated:
> In several dialog-showing functions, the parent parameter is sometimes typed QWidget and sometimes typed MainWindow, and it's sometimes the first parameter and sometimes the last. The API would be nicer if this was made consistent.

jvdwetering:
In Qt, the parent is always the first parameter, so that is probably the way to go. I think in general the parent should be the closest QWidget above it, and so the type should be QWidget.

jvdwetering:
Closing this as it is stale.

PR #158, opened 2023-10-31
Add zxlive.spec file for creating an executable using pyinstaller.

dlyongemallo:
Note: Running this locally will result in an executable containing all the python modules in the virtual environment, which might be very large. For deployment, pyinstaller should be run in a virtual environment with the minimal required libraries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 __pycache__
 build
+dist
 *.egg-info
 .vs
 *.tikz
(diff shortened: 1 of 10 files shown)


dlyongemallo:
FYI: I tested the executable with PR #157 and the executable works as expected: it opens files specified on the command line. However, if a bad file is specified on the command line, issue #156 sometimes happens and the app can be started in a state that looks like it's frozen, because the error message dialog is hidden behind the main window (so the use would have to alt-tab or whatever and close it first before being able to use the app).

PR #173, opened 2023-11-09
Remove dependency on sympy.

dlyongemallo:


RazinShaikh:
I haven't looked at the symbolic part in a while but why do we not need sympy anymore?

dlyongemallo:
Yeah, removing it reduces the size of the executable by a bit, but we still need to really pare down the deployment dependencies.

PR #177, opened 2023-11-10
Open demo graph only if no files specified on command line.

dlyongemallo:


jvdwetering:
This breaks a couple of tests since the tests were assuming that the program doesn't start empty (I think).

dlyongemallo:
Ha ha! Good thing we have tests.

Issue #192, opened 2023-11-16
add more tests


dlyongemallo:
The test coverage is quite limited right now, and it would be good to add more tests. 

In particular, when more functionality is added, tests help both to ensure that regressions or new bugs don't happen and to make it easier for someone reading the code to understand what it is intended to do and to document edge cases or unimplemented features.

A lot of new features were added recently which I'm trying to understand and it would be very helpful to me to see, e.g., what the intended inputs and expected outputs are for typical use cases. (Maybe the inputs/outputs are obvious to someone more versed in ZX-calculus, but I'm still learning.)

RazinShaikh:
We have a lot more tests than we did 3 years ago and while the coverage could still be improved, I don't think we need to keep this issue open as it is a bit open ended.

PR #193, opened 2023-11-16
Clean up some variables not actually used in `MainWindow`

dlyongemallo:
I haven't thoroughly tested the "rewrite" code as I don't understand the typcial workflow, but purely looking at the code, there doesn't seem to be any reason why `rewrite_form`, `left_graph`, and `right_graph` should be owned by `MainWindow` rather than by the only functions in which they're used.


dlyongemallo:
It would be nice if a test were added (see #192) going through a "user journey" for completing a proof using rewrites and so on.

PR #195, opened 2023-11-16
Continuous deployment

RazinShaikh:
`deploy.yml` implements an automatic deployment using `pyside6-deploy` which uses `nuitka`. The final result is a single executable/binary file.
 
`deploy.yml` runs on every new tag pushed with name `v*.*.*` and it automatically creates a new release with the given tag name.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,7 +26,7 @@ jobs:
         run: pip install cx_freeze
       - name: Build MacOS
         if: matrix.os == 'ubuntu-latest'
-        run: python deploy.py bdist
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,9 +1,9 @@
-name: deploy
+name: Deploy and relase
```
jvdwetering, line 1: typo

jvdwetering (commented):
I don't understand this magic, but if you say this works, I say we go for it :)

y-richie-y:
Nice!

PR #206, opened 2023-11-25
Add ability to rename proof steps

### LINKED ISSUE #189: Ability to rename a proof step

wlcsm:
Doubling clicking on the proof step will open a dialog prompting a new name for the proof step.

A separate display name attribute was added to the Rewrite class rather than modifying the "rule" attribute since "rule" is used when converting a proof to a tikz diagram.

This change is backwards compatible since old versions of zxlive with new proof files will ignore the `display_name` attribute and behave as usual, and new versions of zxlive given an old proof file will set the `display_name` to be the name of the rule if it is not present

Fixes #189.

### DIFF
--- a/zxlive/commands.py
+++ b/zxlive/commands.py
@@ -47,6 +47,18 @@ class BaseCommand(QUndoCommand):
         #  we could store "dirty flags" for each node/edge.
         self.graph_view.update_graph(self.g, select_new)
 
+@dataclass
+class UndoableChange(BaseCommand):
+    """ Generic undoable change in the graph that can be appended to the
+    undo stack
+
+    Can be initialised as: UndoableChange(<parent>, <undo_cmd>, <redo_cmd>)
+
+    Where <parent> must contain the graph_view attribute as it is used in
+    BaseCommand.
+    """
+    undo: Callable[[], None]
+    redo: Callable[[], None]
 
 @dataclass
 class SetGraph(BaseCommand):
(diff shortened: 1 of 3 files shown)


RazinShaikh:
Hi Will, can you make renaming rewrite part of the undo-stack? Currently, undoing does not undo the rename. For this, you would need to make renaming a command, which stores the information about the old rewrite. You can find examples of these undo-commands in command.py

RazinShaikh:
Here's an example of how to create it:
https://github.com/Quantomatic/zxlive/blob/fcb1a967d839e14e6f42fd6f3dcde1dfc1e2d818/zxlive/commands.py#L177-L194
and how to call it:
https://github.com/Quantomatic/zxlive/blob/fcb1a967d839e14e6f42fd6f3dcde1dfc1e2d818/zxlive/editor_base_panel.py#L142-L146

wlcsm:
@RazinShaikh I have added it to the undo stack however upon pressing "undo", it doesn't immediately rerender the proof panel which is not a great user experience.

How can I make it rerender the view after the model has been changed? I was looking for a method to call to trigger a rerender, or perhaps I could trigger the undo command via a signal to the ProofModel which I assume will make the proof view rerender.

wlcsm:
@RazinShaikh Fixed the rerendering issue by adding a line to tell the ProofModel which item changed when undoing/redoing https://github.com/Quantomatic/zxlive/pull/206/files#diff-4b24277ddae20b11b98be0dae85205e9f45f931a087edf120fb8203cf59c1555R136

RazinShaikh (approved):

RazinShaikh:
It makes sense that you can just ask the whole panel to render again :)

Issue #207, opened 2023-11-28
Bug saving/loading proofs


wlcsm:
There is a bug where sometimes proofs are not saved or loaded correctly.

Example:

1. Start with the default diagram and fuse the two circled spiders
<img width="351" alt="Screenshot 2023-11-28 at 9 10 23 pm" src="https://github.com/Quantomatic/zxlive/assets/44865292/8f392d8a-af9c-4e30-a2b9-e9b1c8661839">

to get

<img width="341" alt="Screenshot 2023-11-28 at 9 10 28 pm" src="https://github.com/Quantomatic/zxlive/assets/44865292/ffbddba0-f423-4d1d-a82d-c4b3b0997371">

2. Save the proof and open it from the file again to get the following incorrect diagram
<img width="335" alt="Screenshot 2023-11-28 at 9 10 47 pm" src="https://github.com/Quantomatic/zxlive/assets/44865292/664e0549-d20e-4db4-841f-e7cb82031020">

dlyongemallo:
I think this issue is more general than proofs and applies to all graphs. Maybe an issue should be filed named something like "vertices are renumbered on save/open graph, leading to bugs", and *this* specific issue with incorrect proof saving/loading should be considered blocked by that.

As noted in https://github.com/Quantomatic/zxlive/pull/213#issuecomment-1859230170, the numbering of the nodes can change when a graph is saved and then re-opened. 

The numbering of the nodes can matter, which means that the graph which is re-opened is functionally not the same graph as the one which was saved. For example, the "fuse spiders" rule currently fuses the selected vertices into the node with the lowest numbering. When the app is first opened, the nodes are numbered with 0, 1, 2, and 3 as the input nodes, 4, 5, 6, 7 in the second row, and so on. The first image in the first comment in this issue shows the fusion of vertices 14 and 18 by this ordering, which should fuse both into 14 (node on left), as shown here:

![Screenshot from 2023-12-17 22-22-48](https://github.com/Quantomatic/zxlive/assets/9248532/6cc8022a-855e-4eb2-b94d-e83d9dc0fed4|width=350)

This differs from the image shown above, which has fused them into the node on the right. (Presumably, this output was generated after the initial demo graph had been saved, then re-opened.) Of course, since "only connectivity matters", these are really the same diagram. Still, it might be unexpected behaviour for a user to carry out an operation on a graph, and get *visually* different results based on how it was saved/opened.

Similarly, when I re-open the proof, this is what I get:

[...]

Issue #150, opened 2023-10-19
Make a 'only connectivity matters' proof step


jvdwetering:
Right now it is not possible to store movements in the plane as a proof step while in Proof mode. This means when you reload a saved proof that all the information about where the user moved the vertices is lost.

I suggest adding a button like 'save vertex positions as new proof step'.

PR #213, opened 2023-12-16
Remove diffs as internal data structure

wlcsm:
We currently store the proof as both a series of graphs and as an initial graph together with a series of diffs. This PR  removes the diffs to avoid possible synchronisation issues between these two structures.
As a result, it resolves issue #207
    
Additionally, the diffs did not record repositioned vertices, which made adding an OCM step difficult. This then opens up the way for issue #150.

The downside with this change is that it is not backwards compatible with older versions of ZXLive. I've tried to add compatibility by encoding the series of graphs as a series of diffs when saving as JSON. But I encountered problems as wasn't correctly reconstructing the same graph.


wlcsm:
I tried the simple proof where I fuse spiders 0 and 4

Screenshot 2023-12-16 at 9 24 12 pm

But when I save and open the proof, I get

Screenshot 2023-12-16 at 9 24 56 pm

It looks like a problem with computing the diff, since the proof file does contain this extra wire added between node 5 and 8

```
{
  "removed_verts": [
    4
  ],
  "new_verts": [],
  "removed_edges": [],
  "new_edges": [
    [
      0,
      8
    ],
    [.               <------------- Shouldn't be here
      5,
      8
    ]
  ],
  "changed_vertex_types": {},
  "changed_edge_types": {
    "0,8": 1,
    "5,8": 1
  },
  "changed_phases": {},
  "changed_pos": {},
  "changed_vdata": {},
  "variable_types": {}
}
```

wlcsm:
My attempt to encode the proof as a series of diffs is tracked on this branch https://github.com/wlcsm/zxlive/tree/diff-encoding

dlyongemallo:
> So I've not sure how it got there

The vertices are renumbered in `construct_circuit` when the boundary vertices are added in these lines (note the `...+qubits`, where `qubits` is equal to 4):

https://github.com/Quantomatic/zxlive/blob/9822dda6cf5d212c1372db078965d6a6cc16fb21/zxlive/construct.py#L28-L30

https://github.com/Quantomatic/zxlive/blob/9822dda6cf5d212c1372db078965d6a6cc16fb21/zxlive/construct.py#L38-L40

So `4: {5: 1, 8: 1, 0: 1}` is correct, because it corresponds to the edges `(0, 1, 0), (0, 4, 0)` in (the original numbering in) `elist` plus the added edge (from the original vertex `0`, renumbered as `4`) to the added boundary vertex (numbered `0` in the new numbering system), respectively.

I was initially confused by this too when I had to debug this code. I'll add a comment to clarify this.

wlcsm:
> The vertices are renumbered in `construct_circuit` when the boundary vertices are added in these lines (note the `...+qubits`, where `qubits` is equal to 4):

Ah I see, thanks for clarifying.

I've found simply starting a derivation on the demo graph, saving without performing any steps shows the graph state as
```
{0: {4: 1}, 1: {5: 1}, 2: {6: 1}, 3: {7: 1}, 4: {0: 1, 5: 1, 8: 1}, 5: {1: 1, 4: 1, 9: 1, 10: 1}, 6: {2: 1, 10: 1}, 7: {3: 1, 11: 1}, 8: {4: 1, 12: 1}, 9: {5: 1, 13: 2}, 10: {5: 1, 6: 1, 14: 1}, 11: {7: 1, 15: 1}, 12: {8: 1, 16: 1, 17: 1}, 13: {9: 2, 17: 2, 18: 2}, ...}
```

Then after loading, it is different

```
{0: {16: 1, 1: 1, 4: 1}, 1: {17: 1, 0: 1, 5: 1, 6: 1}, 2: {18: 1, 6: 1}, 3: {19: 1, 7: 1}, 4: {0: 1, 8: 1}, 5: {1: 1, 9: 2}, 6: {1: 1, 2: 1, 10: 1}, 7: {3: 1, 11: 1}, 8: {4: 1, 12: 1, 13: 1}, 9: {5: 2, 13: 2, 14: 2}, 10: {6: 1, 13: 1, 14: 1}, 11: {7: 1, 14: 1, 15: 1}, 12: {8: 1, 20: 1}, 13: {8: 1, 10: 1, 21: 1, 9: 2}, ...}
```

Though the graph does load correct which makes me wonder if the vertex indices have been rearranged, and that is cause the diff to be applied incorrectly.

jvdwetering:
I have mixed feelings about not including the diffs in the proof file. I don't have a clear reason for it, but it feels like this will bite us later. Right now the diffs are easily calculated from the graphs themselves, but I could imagine that later we might want to encode some additional information into the diffs (perhaps stuff to do with parametric phases). Anyway, if we remove it now, we could probably also add it back in later in a backwards-compatible way.

wlcsm:
Thats a reasonable concern.
My understanding was that diffs would always contain the same information as the graphs themselves.
Could you explain a little more about what extra information you are considering adding to the diffs?

RazinShaikh (approved):
Hi @wlcsm, I recently returned from holidays and finally had a chance to go through the PR. It all looks good and I am happy to merge this.

RazinShaikh (approved):
Hi @wlcsm, I recently returned from holidays and finally had a chance to go through the PR. It all looks good and I am happy to merge this.

PR #220, opened 2023-12-17
Add OCM step

wlcsm:
Add a button on the toolbar to save the current graph as a separate OCM proof step. Builds upon a previous PR #213
    
The button is always enabled however, even when no changes have been made to the graph. This would be a nice quality of life improvement, but with the way the ProofModel class is set up, it would involve littering button enables/disables throughout which is quite error prone. So I have not included it in this PR.

Screenshot 2023-12-17 at 1 31 43 pm


RazinShaikh:
I haven't tested the PR but could you move the ocm button to the rules panel under the basic rules.

wlcsm:
> I haven't tested the PR but could you move the ocm button to the rules panel under the basic rules.

Done

Screenshot 2023-12-17 at 4 34 10 pm

RazinShaikh (approved):

PR #221, opened 2023-12-17
Enlarge circuit element icons to prevent cropping

### LINKED ISSUE #164: text strings under icons are truncated in side bar
Instead of "Z spider" and "X spider", for example, it says "Z …" and "X …".

sidebar showing truncated text under icons

The list widget's grid size could probably be dynamically set instead of being a fixed width, using `sizeHintForColumn` or something like that.

wlcsm:
The Hadamard wire's name was being cropped as Hadam...

Screenshot 2023-12-17 at 8 42 26 pm

After enlarging the icons by 5 pixels, it becomes

Screenshot 2023-12-17 at 8 42 11 pm

This was to resolve #164, but on my machine all the other circuit element's names were displayed in full. I'm not sure if this is a OS issue (I'm on Mac 14) or it was resolved in a previous commit.

### DIFF
--- a/zxlive/editor_base_panel.py
+++ b/zxlive/editor_base_panel.py
@@ -322,7 +322,6 @@ def create_list_widget(parent: EditorBasePanel,
     list_widget.setViewMode(QListView.ViewMode.IconMode)
     list_widget.setMovement(QListView.Movement.Static)
     list_widget.setUniformItemSizes(True)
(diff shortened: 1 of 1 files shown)


dlyongemallo:
It does vary by OS (and, I think, even by desktop manager and theme), so I think the fix needs to be dynamic.

PR #223, opened 2023-12-21
Add support for Spider QASM as option for Input Circuit.

dlyongemallo:
See https://github.com/Quantomatic/pyzx/issues/192 for context.


jvdwetering:
I was thinking it would be nice to support other circuit formats like QC as well, as it is very simple and easy to input. So maybe we should think ahead and change 'qasm-flavor' to something like 'circuit-format'?

dlyongemallo:
Done.

Issue #218, opened 2023-12-17
crash due to accessing selection after `graph_scene` is deleted


dlyongemallo:
Steps to repro:
1. start a derivation
2. select any node
3. close the app

Expected: app closes

Actual: App crashes in `update_on_selection` in rewrite_action.py, since a selection changed is triggered when the proof panel is closed, and the `graph_scene` is deleted when the app is closed, and the code in `update_on_selection` attempts to use the `graph_scene`. 

It's not such a bad crash since the user is closing the app anyway, but it dumps a stack trace to the console.

```
Traceback (most recent call last):
  File "/home/davinci/workspace/zxlive/zxlive/rewrite_action.py", line 217, in update_on_selection
    selection, edges = self.proof_panel.parse_selection()
  File "/home/davinci/workspace/zxlive/zxlive/proof_panel.py", line 148, in parse_selection
    selection = list(self.graph_scene.selected_vertices)
  File "/home/davinci/workspace/zxlive/zxlive/graphscene.py", line 60, in selected_vertices
    return (it.v for it in self.selectedItems() if isinstance(it, VItem))
RuntimeError: Internal C++ object (GraphScene) already deleted.
```

valleyofblackpanther:
Hi, @dlyongemallo @RazinShaikh to address the crash upon closing the application due to the `update_on_selection` being called after the `graph_scene` has been deleted, we can add a check within the `update_on_selection` method to ensure that the `graph_scene` is still valid. For example:

> `def update_on_selection(self, g, selection, edges) -> None`
> `if self.proof_panel.graph_scene is None or getattr(self.proof_panel.graph_scene, 'g', None) is None:`

### Explanation: 
In the `update_on_selection` method, before iterating over the child items or updating the rewrite, we check if `self.proofpanel.graph_scene` is `None`.  Additionally we check if the `g` attribute of `graph_scene` is present and not `None` If either check fails, the method returns early without performing any updates. This could prevent the application from trying to access a deleted object. Of course, this approach hadn't been tested yet. But if there are any comments I would like to hear.

dlyongemallo:
The best way to verify whether this works or not is to just try it. Please report back on the outcome.

valleyofblackpanther:
Hi @RazinShaikh @dlyongemallo I have made the necessary modifications, this is the [PR 230](https://github.com/Quantomatic/zxlive/pull/230) I did not follow the previous method I mentioned, it didn't work out.

PR #230, opened 2024-01-25
Fixes crash due to accessing selection after graph_scene is deleted #218

valleyofblackpanther:
### Summary 
The `MainWindow `class is where I integrated the cleanup routine for the `GraphScene`. `MainWindow` has already a `closeEvent` method. This is where we call the shutdown method of each `GraphScene` instance before the application closes. However, the `GraphScene` instances are not directly managed by `MainWindow` but rather by the panels like `GraphEditPanel`, `ProofPanel`, and `RulePanel` which are derived from `BasePanel`. Each of these panels contains an instance of `GraphScene`.

Here's the cleanup implementation:
1. Add a cleanup method in the `BasePanel` class.
2. Call this cleanup method for each panel in the MainWindow's closeEvent.

Firslty, I added a `shutdown` method to the `GraphScene` class. Then, added a corresponding method in the `BasePanel` class that calls `GraphScene.shutdown()`
Next, I ensured each panel subclassing `BasePanel` class provides its own shutdown method. 
Finally, I modified the `closeEvent` of the `MainWindow` class to shutdown on each panel before closing. 

This method ensures that each `GraphScene` is properly cleaned up before the main window closes.

### DIFF
--- a/zxlive/dialogs.py
+++ b/zxlive/dialogs.py
@@ -76,9 +76,9 @@ class ImportRuleOutput:
     file_path: str
     r: CustomRule
 
-def show_error_msg(title: str, description: Optional[str] = None) -> None:
+def show_error_msg(title: str, description: Optional[str] = None, parent: Optional[QWidget] = None) -> None:
     """Displays an error message box."""
-    msg = QMessageBox()
+    msg = QMessageBox(parent) # Set the parent of the QMessageBox
     msg.setText(title)
     msg.setIcon(QMessageBox.Icon.Critical)
     if description is not None:
@@ -108,14 +108,14 @@ def create_circuit_dialog(explanation: str, example: str, parent: QWidget) -> Op
     return s if success else None
 
 
-def import_diagram_from_file(file_path: str, selected_filter: str = FileFormat.All.filter) -> \
+def import_diagram_from_file(file_path: str, selected_filter: str = FileFormat.All.filter, parent: Optional[QWidget] = None) -> \
         Optional[ImportGraphOutput | ImportProofOutput | ImportRuleOutput]:
     """Imports a diagram from a given file path.
 
     Returns the imported graph or `None` if the import failed."""
     file = QFile(file_path)
     if not file.open(QIODevice.OpenModeFlag.ReadOnly | QIODevice.OpenModeFlag.Text):
-        show_error_msg(f"Could not open file: {file_path}.")
+        show_error_msg(f"Could not open file: {file_path}.", parent=parent)
         return None
     stream = QTextStream(file)
     data = stream.readAll()
--- a/zxlive/dialogs.py
+++ b/zxlive/dialogs.py
@@ -76,9 +76,9 @@ class ImportRuleOutput:
     file_path: str
     r: CustomRule
 
-def show_error_msg(title: str, description: Optional[str] = None, parent: Optional[QWidget] = None) -> None:
+def show_error_msg(title: str, description: Optional[str] = None) -> None:
     """Displays an error message box."""
-    msg = QMessageBox(parent) # Set the parent of the QMessageBox
+    msg = QMessageBox()
     msg.setText(title)
     msg.setIcon(QMessageBox.Icon.Critical)
     if description is not None:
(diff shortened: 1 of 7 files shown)


zxlive/edit_panel.py
```diff
@@ -106,3 +106,7 @@ def _input_circuit(self) -> None:
             cmd = UpdateGraph(self.graph_view, new_g)
             self.undo_stack.push(cmd)
             self.graph_scene.select_vertices(new_verts)
+
+    def shutdown(self):
```
RazinShaikh, line 110: Why do you need this method here in `edit_panel` if it is already implemented in `base_panel`?

zxlive/proof_panel.py
```diff
@@ -421,6 +421,10 @@ def _refresh_rewrites_model(self) -> None:
         # TODO: Right now this calls for every single vertex selected, even if we select many at the same time
         self.graph_scene.selectionChanged.connect(model.update_on_selection)
 
+    def shutdown(self):
```
RazinShaikh, line 424: Why do you need this method here in `proof_panel` if it is already implemented in `base_panel`?

zxlive/rule_panel.py
```diff
@@ -80,3 +80,7 @@ def get_rule(self) -> CustomRule:
                           self.graph_scene_right.g,
                           self.name_field.text(),
                           self.description_field.text())
+    
+    def shutdown(self):
```
RazinShaikh, line 84: Why do you need this method here in `rule_panel` if it is already implemented in `base_panel`?

zxlive/graphscene.py
```diff
@@ -195,6 +195,20 @@ def select_all(self) -> None:
...
+    def shutdown(self):
+
+        """
+        Perform necessary cleanup to avoid RunTimeError when the application closes.
+        Call this method before the application closes. 
+        """        
+
+        #Clear all items from the scene to ensure no dangling references remain
+        self.clear()
+
```
RazinShaikh, line 207: Do you mind cleaning up some extra empty lines in this function?

zxlive/dialogs.py
```diff
@@ -76,9 +76,9 @@ class ImportRuleOutput:
     file_path: str
     r: CustomRule
 
-def show_error_msg(title: str, description: Optional[str] = None) -> None:
+def show_error_msg(title: str, description: Optional[str] = None, parent: Optional[QWidget] = None) -> None:
```
RazinShaikh, line 79: The changes in this file regarding the parent are related to a separate issue. Do you mind splitting this into a separate PR?

RazinShaikh (changes requested):
Hi Kaushik, thank you for the PR. Could you please make the requested changes before we can merge this?

zxlive/graphscene.py
valleyofblackpanther: Hi, @RazinShaikh I have made the suggested changes, here is the [commit](https://github.com/Quantomatic/zxlive/pull/230/commits/cb113bde6d0ba31fad648426a89c990a0b4a8d2f)

valleyofblackpanther:
Hi, @RazinShaikh I have made the suggested changes, here is the [commit](https://github.com/Quantomatic/zxlive/pull/230/commits/cb113bde6d0ba31fad648426a89c990a0b4a8d2f).

dlyongemallo:
I wonder if it wouldn't be simpler and more future-proof to simply clear the node selection when the user asks to close the app?

RazinShaikh:
@dlyongemallo that actually makes more sense

zxlive/mainwindow.py
```diff
@@ -223,6 +223,12 @@ def active_panel(self) -> Optional[BasePanel]:
 
 
     def closeEvent(self, e: QCloseEvent) -> None:
+        #Call shutdown on all tabs before closing
```
dlyongemallo, line 226: I think you may be overthinking this. I think all that is needed to fix the issue is to add a single line in `close_tab` (just above line 329 `self.tab_widget.removeTab(i)`), like:
`widget.graph_scene.clearSelection()`

Please try to see if that works.

dlyongemallo:
> Hi Kaushik, thank you for the PR. Could you please make the requested changes before we can merge this?

Thanks for the PR, Kaushik. Can you make the suggested change and add a test to confirm it works? I think that should suffice.

PR #240, opened 2024-05-05
Move UI code for showing errors out of `custom_rule.py`.

dlyongemallo:
Incidentally, make the warning dialogs modal (see issue #156).

### DIFF
--- a/zxlive/custom_rule.py
+++ b/zxlive/custom_rule.py
@@ -315,30 +315,28 @@ def get_vertex_positions(graph: GraphT, rhs_graph: nx.Graph, boundary_vertex_map
     ret: dict[NodeView, tuple[float, float]] = nx.spring_layout(rhs_graph, k=k, pos=pos_dict, fixed=boundary_vertex_map.keys())
     return ret
 
+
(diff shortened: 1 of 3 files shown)


dlyongemallo:
Two further questions:
* `check_rule` returns a `bool`, but this is never used. Is this return value necessary?
* `check_rule` is always called with `show_error = True`. Is this parameter necessary?

jvdwetering:
I think the bool was originally intended to check if the operation succeeded. But since with this commit it instead raises an Exception, that no longer makes sense. Also with this behaviour, supressing the exceptions with show_error also makes less sense so that that parameter can be removed.

dlyongemallo:
Okay, I've removed both the parameter and the return value.

hm919:
Has this fixed issue #156 ?

dlyongemallo:
The specific issue of `show_error_msg` not taking a parent has been fixed. It looks to me like all dialog functions defined in `dialogs.py` take a parent now. The aside about the API being inconsistent is still true, but that's mostly a cosmetic matter and I think we can close issue #156 now (or leave it open and rename it to "make dialogs more API consistent" or something like that).

Issue #199, opened 2023-11-21
Label boundary nodes with input/output index so that it is more clear to the user, especially when creating a custom rewrite 


RazinShaikh:
This should be the index of the boundary node in graph.inputs and graph.outputs

PR #246, opened 2024-06-12
label boundary nodes in a custom rule

RazinShaikh:


RazinShaikh:
Closes https://github.com/Quantomatic/zxlive/issues/199

PR #251, opened 2024-06-13
Add sound effects

y-richie-y:
Sound effects are activated with the shortcut Ctrl+B.


RazinShaikh:
Epic 😎 Can you add an option to turn the sound effects on or off in the menu or preferences? Another slight problem I noticed is that when I have the sound effects on, the UI gets stuck for ~0.5 seconds to play the sound effect. Can you please fix that?

y-richie-y:
Added the sound effects in preferences. On my computer the lag only happens the first time any sound is played, not sure what to do about that.

RazinShaikh:
The lag happens every time for me unfortunately.

y-richie-y:
Does anyone have any ideas on how to fix this? I can't replicate it on my machine

RazinShaikh:
After the recent commit, I am not able to open preferences. I get the following error:
```
Traceback (most recent call last):
  File "C:\Users\Razin Shaikh\OneDrive\Documents\code\zxlive\zxlive\mainwindow.py", line 129, in <lambda>
    self.preferences_action = self._new_action("&Preferences...", lambda: open_settings_dialog(self), None, "Open the preferences dialog")
  File "C:\Users\Razin Shaikh\OneDrive\Documents\code\zxlive\zxlive\settings_dialog.py", line 264, in open_settings_dialog
    dialog = SettingsDialog(parent)
  File "C:\Users\Razin Shaikh\OneDrive\Documents\code\zxlive\zxlive\settings_dialog.py", line 67, in __init__
    self.add_setting(form_general, "sound-effects", "Sound Effects", "checkbox")
  File "C:\Users\Razin Shaikh\OneDrive\Documents\code\zxlive\zxlive\settings_dialog.py", line 225, in add_setting
    widget.setChecked(val)
TypeError: 'PySide6.QtWidgets.QAbstractButton.setChecked' called with wrong argument types:       
  PySide6.QtWidgets.QAbstractButton.setChecked(str)
Supported signatures:
  PySide6.QtWidgets.QAbstractButton.setChecked(bool)
```

y-richie-y:
I don't get this error but I have (hopefully) a fix

jvdwetering:
In Qt, you can add files to a resource file, and then they will be loaded at start time. This might prevent the lagging when playing the sound for the first time (but at a potential cost of increasing the start-up time).
If you want to try this, you can add the sound files to the zxlive/resources.qrc file. You then have to recompile the file to resources.py: https://www.pythonguis.com/tutorials/packaging-data-files-pyside6-with-qresource-system/

RazinShaikh:
Perhaps we should just merge this as this is an easter egg feature, disabled by default. Thoughts?

y-richie-y:
I am for this: I can't replicate the lag on my machine and I also don't know what could fix the lag

jvdwetering:
The testing suite gives the error
```
from PySide6.QtMultimedia import QSoundEffect
E   ImportError: libpulse.so.0: cannot open shared object file: No such file or directory
```
Should a new library be included in the requirements.txt?

y-richie-y:
LGTM! ✌️

jvdwetering:
The boolean settings is actually broken: if you click the checkmark and press "Apply" in the settings dialog, and then open it again, the checkmark is again empty. But it seems that regardless the sound effects are on.

PR #252, opened 2024-06-13
Multigraph support in edit mode

RazinShaikh:
I would like to split multigraph support work into several PRs and I think it is fine to merge as the main branch is already broken. This PR adds the ability to create multiple edges in the edit mode with multigraph. This PR depends on the PR https://github.com/Quantomatic/pyzx/pull/236 in pyzx which should be merged first.

### DIFF
--- a/zxlive/graphscene.py
+++ b/zxlive/graphscene.py
@@ -108,8 +108,9 @@ class GraphScene(QGraphicsScene):
                 self.removeItem(v_item.phase_item)
             for anim in v_item.active_animations.copy():
                 anim.stop()
-            for e in self.g.incident_edges(v):
-                removed_edges.add(e)
+            for s, t in self.g.incident_edges(v):
+                for e in self.g.edges(s, t):
+                    removed_edges.add(e)
 
             selected_vertices.discard(v)
             self.removeItem(v_item)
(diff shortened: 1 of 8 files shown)


zxlive/graphscene.py
```diff
@@ -52,7 +52,7 @@ def __init__(self) -> None:
         self.setSceneRect(0, 0, 2*OFFSET_X, 2*OFFSET_Y)
         self.setBackgroundBrush(QBrush(QColor(255, 255, 255)))
         self.vertex_map: dict[VT, VItem] = {}
-        self.edge_map: dict[ET, EItem] = {}
+        self.edge_map: dict[tuple[ET, int], EItem] = {}
```
jvdwetering, line 55: Is this correct? You already redefined ET to have an additional int in it.

jvdwetering:
This is giving 'AttributeError: 'Edge' object has no attribute 'get_edge_count'' in the test suite. Is this expected?

zxlive/graphscene.py
RazinShaikh: The type of ET has (source, target, edge_type) but this is not enough information to determine which exact edge it is in the drawing. So I give an additional index in the edge_map.

RazinShaikh:
I added the get_edge_count method in the PyZX PR that's now merged.
https://github.com/Quantomatic/pyzx/blob/a11c8f8a2d6a086a0b9a2cb69ab85384a1aa762a/pyzx/graph/multigraph.py#L53-L57

PR #254, opened 2024-06-21
Fix image scaling in documentation

wlcsm:
With the "scaling" attribute, images and gifs are assigned a fixed width and heigh at compile time which is included directly in the html - overwriting the rest of the style sheets which tell it to auto scale.

This caused images to be squished on mobile devices.

Before

After 
Screenshot 2024-06-21 at 10 05 52 AM


boldar99 (approved):
Works for me as well on chrome and safari

PR #255, opened 2024-06-21
Maintain sizes of splitter between tabs

### LINKED ISSUE #219: UI suggestion: proof panel splitter should maintain setting between tabs
If the user opens multiple proof panels, the splitters (the vertical bars separating the rules, the graph, and the proof steps) are set independently in each tab. When tabbing between different proofs, this has the effect of moving the UI around unexpectedly. 

For example, I often widen the rules widget to see the names of longer rules, but when I tab to another proof, the names are cut off again. So I have to manually widen the rules widget every time I start a new proof, which is annoying. (Maybe the widths of the rules and proof steps widgets can even be saved as a setting.)

boldar99:
Closes #219

### DIFF
--- a/zxlive/base_panel.py
+++ b/zxlive/base_panel.py
@@ -113,5 +113,5 @@ class BasePanel(QWidget):
         self.splitter_sizes[self.__class__] = self.splitter.sizes()
 
     def set_splitter_size(self):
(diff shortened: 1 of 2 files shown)


RazinShaikh (approved):
looks good

PR #282, opened 2024-07-03
improvement(typing): Use PySide6 generated stubs

boldar99:
Since PySide6>=6.7 includes stub files, we can drop the PySide6-stubs dependency.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,8 +24,9 @@ jobs:
         cache: 'pip'
     - name: Install dependecies
       run: |
+        sudo apt-get install -y libegl1
         pip install ".[test]"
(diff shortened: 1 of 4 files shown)


jvdwetering:
This page says that pyside6 already ships with the stubs: https://doc.qt.io/qtforpython-6/tools/pyside-genpyi.html
Is it necessary for us to regenerate them?

boldar99:
Yeah, you're right, I missed that. I'll fix these soon. Should I keep the changes to README.md (excluding the part about `pyside6-genpyi`) or revert the file complately?

RazinShaikh:
Why not keep the info about running tests in the readme? It's good to have and is only a few sentences.

PR #293, opened 2024-07-05
Making rewrite_action.py well-typed

boldar99:


RazinShaikh (approved):
Nice, I was looking into this last night but couldn't figure out how to fix the type errors in rewrite_action.py

PR #294, opened 2024-07-07
Refactoring settings

boldar99:
Most information related to settings has been stored in `common.py`, and the implementation of `settings_dialog.py` was not the cleanest.

This PR intends to:
1. make the settings related code type-safe,
2. move the information related to settings to a new `settings.py` file, and
3. clean up the implementations related to settings. 

Most changes are simply moving data to settings.py and extracting methods in settings_dialoge.py (which are sometimes further abstracted to reduce code repetition). Furthermore, some `TypedDict`s are defined to standardise the data in `settings.py`.


RazinShaikh:
This looks like a lot of changes. Can you explain what's going on at the high level?

zxlive/settings.py
```diff
@@ -0,0 +1,204 @@
...
+    "tikz/names/copy 0/pi spider": "cp",
+    "tikz/names/push Pauli": "pi",
+    "tikz/names/decompose hadamard": "eu",
+}
+
+defaults = general_defaults | tikz_export_defaults | tikz_import_defaults \
+           | tikz_layout_defaults | tikz_names_defaults
+
+
+moder_red_green: ColorsScheme = {
```
jvdwetering, line 89: typo? moder vs modern

zxlive/settings_dialog.py
```diff
@@ -15,167 +15,164 @@
...
+}
+
+tab_positioning_data = {
+    QTabWidget.TabPosition.North: "Top",
+    QTabWidget.TabPosition.South: "Bottom",
+    QTabWidget.TabPosition.West: "Left",
+    QTabWidget.TabPosition.East: "Right"
+}
+
+snap_to_grpid_data = {'2': "2", '4': "4", '8': "8", '16': "16"}
```
jvdwetering, line 64: grpid -> grid?

zxlive/settings_dialog.py
```diff
@@ -15,167 +15,164 @@
...
 
-        ##### Tikz Layout settings #####
-        panel_tikz_layout = QWidget()
-        vlayout = QVBoxLayout()
-        panel_tikz_layout.setLayout(vlayout)
-        tab_widget.addTab(panel_tikz_layout, "Tikz layout")
+        self.add_settings_tab(tab_widget, "Tikz layout",  "Tikz layout settings", tikz_layout_settings)
 
-        vlayout.addWidget(QLabel("Tikz layout settings"))
+        self.init_okay_cancle_buttons(layout)
```
jvdwetering, line 149: cancle -> cancel?

jvdwetering:
It looks like Boldi is making my janky implementation less spaghetti-like, by using named dictionary types, and spreading out some of the data from the definition of the Settings class. This is probably a good idea.

zxlive/settings_dialog.py
```diff
@@ -187,49 +184,75 @@ def __init__(self, main_window: MainWindow) -> None:
...
+                setattr(widget, "text_value", directory)
+
+        hlayout.addWidget(widget_line)
+        button = QPushButton("Browse")
+        button.clicked.connect(browse)
+        hlayout.addWidget(button)
+        return widget
+
+    def make_combo_form_input(self, data: SettingsData) -> QComboBox:
+        name, _data = data["id"], data["data"]
```
RazinShaikh, line 227: Should you check if the key `data["data"]` exists as it is not required in the dictionary?

boldar99:
I have updated the description of the PR, intending to explain what's going on.

zxlive/settings_dialog.py
```diff
@@ -187,49 +184,75 @@ def __init__(self, main_window: MainWindow) -> None:
...
+        value = self.settings.value(name, defaults[name])
+        widget = QComboBox()
+        widget.addItems(list(_data.values()))
+        widget.setCurrentText(_data[value])
+        setattr(widget, "data", _data)
```
RazinShaikh, line 232: It might be better to subclass QComboBox instead of dynamically adding the `data` attribute.

zxlive/settings_dialog.py
```diff
@@ -187,49 +184,75 @@ def __init__(self, main_window: MainWindow) -> None:
...
+        widget_line = QLineEdit()
+        widget_line.setText(value)
+
+        def browse() -> None:
+            directory = QFileDialog.getExistingDirectory(
+                self,"Pick folder", options=QFileDialog.Option.ShowDirsOnly
+            )
+            if directory:
+                widget_line.setText(directory)
+                setattr(widget, "text_value", directory)
```
RazinShaikh, line 218: same here, avoid dynamically setting attributes

zxlive/common.py
```diff
@@ -1,22 +1,34 @@
...
 
 
+T = TypeVar('T')
+
+
 def get_data(path: str) -> str:
     return os.path.join(os.environ.get("_MEIPASS", _ROOT), path)
 
+
+def get_settings_value(arg: str, _type: Type[T], default: T | None = None, settings: QSettings | None = None) -> T:
```
RazinShaikh, line 24: I was looking up what `Type[T]` meant as I was not familiar with it. I found that `typing.Type` is depricated as the builtin `type` now supports indexing. See https://docs.python.org/3/library/typing.html#typing.Type

boldar99:
Okay, I have modified the implementation to use built-in methods instead of tricks with `setattr`.

RazinShaikh (approved):
nice, shall we merge it then?

zxlive/settings_dialog.py
```diff
@@ -15,168 +15,160 @@
...
-            QTabWidget.TabPosition.North: "Top", QTabWidget.TabPosition.South: "Bottom",
-            QTabWidget.TabPosition.West: "Left", QTabWidget.TabPosition.East: "Right"})
-        self.add_setting(form_general, "snap-granularity", "Snap-to-grid granularity", 'combo',
-                         data = {'2': "2", '4': "4", '8': "8", '16': "16"})
-        self.add_setting(form_general, "input-circuit-format", "Input Circuit as", 'combo', data=input_circuit_formats)
-        self.prev_color_scheme = self.settings.value("color-scheme")
-        self.prev_tab_bar_location = self.settings.value("tab-bar-location")
-        vlayout.addStretch()
+        self.add_settings_tab(tab_widget, "General", "Tikz rule name settings", general_settings)
+        self.add_settings_tab(tab_widget, "Tikz rule names", "General ZXLive settings", tikz_rule_name_settings)
```
RazinShaikh, line 142: Sorry, I just noticed that the name and description are flipped here

RazinShaikh: ok i have fixed it

PR #300, opened 2024-07-09
Font settings

### LINKED ISSUE #278: Increase font size of the ZXLive UI?
Is it just me or do other people also find the default font size of the ZXLive UI a bit too small?

boldar99:
This PR enables setting the font size and family to be used across the application.

This is achieved mainly via `QApplication.setFont(...)`, but some elements don't change with that, so those are set manually in `update_font` methods. Also, some previously hard-coded fonts were removed.

The current font size and family are accessible in `zxlive.settings.display_setting.font: QFont`, and users can update it in a new "Font" tab of the preferences window.

Closes #278

### DIFF
--- a/zxlive/settings.py
+++ b/zxlive/settings.py
@@ -4,7 +4,7 @@ from typing import Dict, Any, TypedDict
 
 import pyzx
 from PySide6.QtCore import QSettings
-from PySide6.QtGui import QColor
+from PySide6.QtGui import QColor, QFont
 from PySide6.QtWidgets import QTabWidget
 
 from .common import get_settings_value, SCALE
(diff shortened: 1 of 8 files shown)


RazinShaikh:
The font size of the menu bar does not update on windows until the app is restarted

zxlive/settings.py
```diff
@@ -28,14 +28,19 @@ class ColorScheme(TypedDict):
...
+general_defaults: dict[str, str | QTabWidget.TabPosition | int] = {
     "path/custom-rules": "lemmas/",
     "color-scheme": "modern-red-green",
     "tab-bar-location": QTabWidget.TabPosition.North,
     "snap-granularity": '4',
     "input-circuit-format": 'openqasm',
 }
 
+font_defaults: dict[str, str | int | None] = {
+    "font/size": 13,
```
RazinShaikh, line 40: I would suggest 11 as the default size. QT's default is 9, which is small but 13 might be a bit too large. Let me know what looks good on your computer.

boldar99:
Fixed!

PR #305, opened 2024-07-10
Detect inputs and outputs before saving as JSON

zickgraf:
JSON explicitly stores the inputs and outputs and their order (I'm not sure about the other formats). Having this information available in the saved file is useful for reading the JSON with other programs.

### DIFF
--- a/zxlive/dialogs.py
+++ b/zxlive/dialogs.py
@@ -210,6 +210,10 @@ def save_diagram_dialog(graph: GraphT, parent: QWidget) -> Optional[tuple[str, F
     file_path, selected_format = file_path_and_format
 
     if selected_format in (FileFormat.QGraph, FileFormat.Json):
+        try:
+            graph.auto_detect_io()
+        except TypeError:
+            pass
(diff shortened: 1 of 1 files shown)


zxlive/dialogs.py
```diff
@@ -210,6 +210,7 @@ def save_diagram_dialog(graph: GraphT, parent: QWidget) -> Optional[tuple[str, F
     file_path, selected_format = file_path_and_format
 
     if selected_format in (FileFormat.QGraph, FileFormat.Json):
+        graph.auto_detect_io()
```
RazinShaikh, line 213: graph.auto_detect_io() can sometimes fail. Can you put it inside a try block so that the save works even when detecting io failed?
```python
try:
    graph.auto_detect_io()
except TypeError:
    pass
```

RazinShaikh (changes requested):

zxlive/dialogs.py
zickgraf: Ah I see, I have updated the PR :-)

PR #309, opened 2024-07-10
Add preview pictures to rewrites

jvdwetering:
Made it so when you hover over a rewrite in the big list, it shows a preview picture. This also works for custom rewrites. These are just pictures and not gifs, as that was too hard to do.
A couple of things I leave open for now:

1. Currently I've manually made pictures based on papers that show the rewrite. The benefit of this is that this can show stuff like 'dot-dot-dots' to denote more symbolically what the rewrite does. The drawback is that if the user changes the colour style of the diagrams, that the previouw picture does not match what is shown in ZXLive. A solution to this would be to automatically generate a preview picture based on what is in ZXLive, but this is very high effort.
2. We need to decide whether we want every rewrite to have an associated picture, or only some that are potentially more difficult to understand.


RazinShaikh:
Can you add an option to enable or disable previews in settings?

RazinShaikh:
Regarding 3, previews for custom rewrites are quite necessary. I often forget what the rewrite is supposed to do and then I need to open the actual file to check.

jvdwetering:
I made it a toggleable setting, but it only applies to a newly opened proof tab, but that's probably fine.

zxlive/rewrite_data.py
```diff
@@ -57,15 +56,17 @@ def read_custom_rules() -> list[RewriteData]:
         "matcher": pyzx.rules.match_lcomp_parallel,
         "rule": pyzx.rules.lcomp,
         "type": MATCHES_VERTICES,
-        "copy_first": True
+        "copy_first": True,
+        "picture": "lcomp.png"
```
boldar99, line 60: You should add `picture` to RewriteData

boldar99:
The files in the tooltips folder need to be added to `MANIFEST.in` so that those files are also distributed with pip.

zxlive/resources.py
RazinShaikh, line 1: Just wondering if this file with 41,595 lines should be pushed to the repo. It looks like it was auto-generated.

zxlive/settings_dialog.py
```diff
@@ -71,13 +71,19 @@ class SettingsData(TypedDict):
...
+}
+
 
 general_settings: list[SettingsData] = [
     {"id": "path/custom-rules", "label": "Custom rules path", "type": FormInputType.Folder},
     {"id": "color-scheme", "label": "Color scheme", "type": FormInputType.Combo, "data": color_scheme_data},
     {"id": "tab-bar-location", "label": "Tab bar location", "type": FormInputType.Combo, "data": tab_positioning_data},
     {"id": "snap-granularity", "label": "Snap-to-grid granularity", "type": FormInputType.Combo, "data": snap_to_grid_data},
     {"id": "input-circuit-format", "label": "Input Circuit as", "type": FormInputType.Combo, "data": input_circuit_formats},
+    {"id": "previews-show", "label": "Show rewrite previews","type": FormInputType.Combo, "data": combo_true_false}
```
RazinShaikh, line 86: this will look better as a checkbox instead of a dropdown with true or false

boldar99: Or potentially Toggle switch, but I think it's a great issue for someone during the hackathon: #321

PR #324, opened 2024-07-12
Rename Quantomatic links to zxcalc.

dlyongemallo:


boldar99:
Maybe the value of `myappid` (app.py:36) can also be changed to use zxlive instead of quantomatic.

dlyongemallo:
> Maybe the value of `myappid` can also be changed to use zxcalc instead of quantomatic.
> 
> https://github.com/dlyongemallo/zxlive/blob/9dbe1bb1203c899cd49bb014be328961658098fb/zxlive/app.py#L35

Done.

boldar99 (approved):
LGTM

PR #329, opened 2024-07-17
Add export to gif

mjsutcliffe99:
Add option, under file tab, to export proof as gif


y-richie-y:
Looks good but you should add `imageio` to `requirements.txt`

Issue #321, opened 2024-07-11
Add boolean settings form input


boldar99:
Potentially with a toggle box or toggle switch.

Steps:

1. Add `Bool` to `FormInputType`
2. Implement `SettingsDialog.make_bool_form_input`
3. Add it to `add_setting_to_form`
4. Change all the boolean settings to use the new implementation.
5. Add hidden settings like sparkle mode and sound effects (If done)

jvdwetering:
This is implemented, but currently broken: the settings aren't actually saved.

PR #330, opened 2024-07-17
Fix all bool settings

y-richie-y:
The show preview setting only updates upon restart as they are generated on start up


y-richie-y:
Closes #321

boldar99 (approved):
Looks good to me!

Issue #311, opened 2024-07-10
Command to snap spiders on to the wire


RazinShaikh:
Suppose we have a lot of phase gadgets and we want to connect the top row of spiders together through a wire. We then need to connect each spider to the next threading our way from left to right. This is very tedious. It would be useful to have a shortcut which can connect the selected spiders to the wire below it.
![image](https://github.com/zxcalc/zxlive/assets/28257517/7f088bb2-c788-489d-a3ef-1e9f0baa217a)

jvdwetering:
Implemented by #337

Issue #302, opened 2024-07-09
Edge should be selectable only by clicking directly on the edge, not simply by clicking in the interior region


RazinShaikh:
An edge can be selected and dragged by clicking inside the interior region as shown below:

https://github.com/zxcalc/zxlive/assets/28257517/cbaea426-7efe-457b-a6c3-403063d422d0

This is undesirable because it causes this weird layering of regions where we can't select an edge because it is hidden under the region of a different edge:

https://github.com/zxcalc/zxlive/assets/28257517/3e3f0b36-6114-4434-9694-205ca2aaf638

Moreover, we can't use magic wand in the interior region.

RazinShaikh:
Same issue here: https://forum.qt.io/topic/118701/qpainterpath-selection-problem

jvdwetering:
Tikzit is also written in qt and does this properly. So probably @akissinger knows a solution

PR #337, opened 2024-07-22
Add toggle to automatically connect new vertex to edge underneath it

jvdwetering:
Adds a button next to the Selection/Vertex/Edge buttons which when toggled makes the 'add vertex' automatically connect on an edge underneath it.

To make clear it snapped I added an animation to the edge, for which I added a new EItemAnimation class.


jvdwetering:
This is related to #311, but then kind of the converse.

RazinShaikh:
This is very hit or miss. Can you do something about that?

https://github.com/user-attachments/assets/8aaf2664-3b39-46c2-80cc-e4cd2ab7f11a

jvdwetering:
I've mostly been dealing with this by zooming in more, so that the edge is wider. An alternative solution I was thinking of, was to make an edge you are hovering over more thick so that it is easier to click on it.
There is also the manual solution, where we don't use the built-in Qt collision features, but instead check ourselves across all edges whether there is an intersection (or something close to it), this might get slow for large diagrams though.
I should also add that the current implementation suffers from the same issue described in #302.

zxlive/graphscene.py
```diff
@@ -291,7 +293,9 @@ def mouseReleaseEvent(self, e: QGraphicsSceneMouseEvent) -> None:
 
     def add_vertex(self, e: QGraphicsSceneMouseEvent) -> None:
         p = e.scenePos()
-        self.vertex_added.emit(*pos_from_view(p.x(), p.y()))
+        # edges under current mouse position
+        edges: list[EItem] = [e for e in self.items(p, deviceTransform=QTransform()) if isinstance(e,EItem)]
```
RazinShaikh, line 297: I figured out how to do check for edge intersection by creating a rectangle around the mouse position. This can be based on the grid size. I will push that change.

zxlive/eitem.py
```diff
@@ -197,3 +211,70 @@ def compute_perpendicular_direction(source_pos: QPointF, target_pos: QPointF) ->
...
+
+    e: Optional[ET]
+
+    def __init__(self, item: Union[EItem, ET], property: EItem.Properties,
+                 scene: Optional[GraphScene] = None, refresh: bool = False) -> None:
+        super().__init__()
+        self.e = None
+        self._it = None
+        self.scene: Optional[GraphScene] = None
+        #if refresh and property != VItem.Properties.Position:
```
RazinShaikh, line 236: remove the comment?

zxlive/eitem.py
```diff
@@ -197,3 +211,70 @@ def compute_perpendicular_direction(source_pos: QPointF, target_pos: QPointF) ->
     direction = direction / norm
     perpendicular = QPointF(-direction.y(), direction.x())
     return perpendicular
+
+
+class EItemAnimation(QVariantAnimation):
```
RazinShaikh, line 216: This class and the `VItemAnimation` class shares most of the code. It will be cleaner to factor out the common code in a parent class.

RazinShaikh (changes requested):

zxlive/graphscene.py
RazinShaikh: Pushed now!

zxlive/editor_base_panel.py
```diff
@@ -313,7 +342,16 @@ def toolbar_select_node_edge(parent: EditorBasePanel) -> ToolbarSection:
     select.clicked.connect(lambda: parent._tool_clicked(ToolType.SELECT))
     vertex.clicked.connect(lambda: parent._tool_clicked(ToolType.VERTEX))
     edge.clicked.connect(lambda: parent._tool_clicked(ToolType.EDGE))
-    return ToolbarSection(select, vertex, edge, exclusive=True)
+    yield ToolbarSection(select, vertex, edge, exclusive=True)
+
+    snap = QToolButton(parent)
+    snap.setCheckable(True)
+    snap.setChecked(True)
+    snap.setText("Vertex-edge snap")
```
RazinShaikh, line 350: Not necessary for this PR but it would be nice to have a sleek icon instead of the text button

zxlive/rewrite_action.py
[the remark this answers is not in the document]
jvdwetering: I don't understand this request

zxlive/eitem.py
jvdwetering: While that's true, quite a lot of it actually can't be made common without doing some generic typing that would be harder to interpret as well. This is because the EItem and VItem don't have useful common ancestor and the properties we can animate are different for each of these.

zxlive/editor_base_panel.py
jvdwetering: I added an icon which I think looks pretty neat :)
(courtesy of chatGPT with a dash of my own creative inspiration)

jvdwetering:
This now also implements #311.
I think this is ready for merging

zxlive/rewrite_action.py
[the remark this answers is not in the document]
RazinShaikh: image

one of your commits (probably accidently) added a whitespace and a blank line. I was asking if you can undo that. It's not super important though.

zxlive/eitem.py
RazinShaikh: that's fine, we can leave it to future work

zxlive/editor_base_panel.py
```diff
@@ -347,8 +374,9 @@ def toolbar_select_node_edge(parent: EditorBasePanel) -> Iterator[ToolbarSection
     snap = QToolButton(parent)
     snap.setCheckable(True)
     snap.setChecked(True)
-    snap.setText("Vertex-edge snap")
-    snap.setToolTip("Snap newly added vertex to the edge beneath it (f)")
+    #snap.setText("Vertex-edge snap")
```
RazinShaikh, line 377: remove comment

zxlive/graphscene.py
```diff
@@ -293,14 +294,31 @@ def mouseReleaseEvent(self, e: QGraphicsSceneMouseEvent) -> None:
...
         self.removeItem(self._drag)
+        v1 = self._drag.start
+        self._drag = None
         for it in self.items(e.scenePos(), deviceTransform=QTransform()):
             if isinstance(it, VItem):
-                self.edge_added.emit(self._drag.start.v, it.v)
-        self._drag = None
+                v2 = it
+                break
+                #self.edge_added.emit(self._drag.start.v, it.v)
```
RazinShaikh, line 313: remove comment

RazinShaikh (approved):
looks good to me

jvdwetering:
This is giving mypy a hard time:
```
def it(self) -> EItem:
        if self._it is None and self.scene is not None and self.e is not None:
            self._it = self.scene.edge_map[self.e]
        assert self._it is not None
        return self._it
```
we have `_it: Optional[EItem]`.
The error is ` Incompatible types in assignment (expression has type "Dict[int, EItem]", variable has type "Optional[EItem]") `.
I don't understand why it is an error to assign a `EItem` to something of type `Optional[EItem]`.

RazinShaikh:
Hi @jvdwetering , thanks for merging this. For the next time, do you mind merging the PRs without squashing the commits. I think for a big PR like this, it is quite useful to have the full commit history for future reference/debugging instead of a single commit containing all the changes.

jvdwetering:
Okay that makes sense. I wasn't sure what a good policy would be.

PR #339, opened 2024-07-22
Made previews only load once you first hover over the rewrite rule, cutting down on loading time

jvdwetering:


RazinShaikh (approved):
It looks good to me. Shall we merge it?

RazinShaikh:
This PR broke the previews settings. The tooltip shows the preview even when the "Show rewrite previews" is turned off.

RazinShaikh:
We need to use 
```python
if display_setting.previews_show
```
but I think it might need a few more changes because currently the following line overwrites the tooltip_str.
https://github.com/zxcalc/zxlive/blob/40d0ea0db725c584ccd0d8bcd1cbc5a82fe87f4d/zxlive/rewrite_action.py#L156

jvdwetering:
Fixed in f1fb7b247a31a2a8d621f1458faa34ca246ceadc

PR #344, opened 2024-07-29
Move "refresh rules" in a context menu

RazinShaikh:
Removed the refresh rules button and added a context menu on right click to perform that action


boldar99 (approved):
Nice!

PR #345, opened 2024-07-29
Preserve selection across undo/redo

### LINKED ISSUE #186: maintain selection state across undo/redo

RazinShaikh:
I am not sure if this is good or bad for the user experience. So I would like a few people to give feedback.

Closes #186


jvdwetering:
I am a fan of this. 
Also, it mimicks the behaviour of Tikzit, which is what we are always trying to do.

RazinShaikh:
okay, merging this then :)

PR #360, opened 2024-10-23
Fix ZXLive not loading graphs as Multigraphs with auto simplify disabled

jvdwetering:
When you load a graph, it was sometimes loaded as a GraphS instead of a Multigraph, and in all cases set_auto_simplify was not set properly leading to unexpected behaviour when adding edges (and exceptions).


jvdwetering:
I'm merging this, but I think the solution is quite ugly, and error prone. Ideally we would set some global default value that Multigraph does not auto simplify everywhere.

PR #371, opened 2025-05-20
Add matrix precision control and clipboard copy to matrix dialog

### LINKED ISSUE #316: Set precision of matrix and copy it to clipboard
It would be nice to set the precision of the displayed matrix either in settings or in the dialogue.

Another nice feature would be if the matrix could be copied to the clipboard with a click of a button as you sometimes want to play around with the matrix.

RazinShaikh:
Fixes #316 

- Add 'Matrix display precision' to settings dialog
- Use setting as default in matrix view
- Add precision spinbox and 'Copy to Clipboard' button to matrix dialog
- Refactor for mypy and code clarity

### DIFF
--- a/zxlive/mainwindow.py
+++ b/zxlive/mainwindow.py
@@ -560,14 +560,20 @@ class MainWindow(QMainWindow):
         self.active_panel.graph_view.fit_view()
 
     def show_matrix(self) -> None:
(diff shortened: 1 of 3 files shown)


RazinShaikh:
This PR was implemented using github copilot agent mode

Issue #268, opened 2024-06-29
Support for blank nodes to add text, maths, etc.


RazinShaikh:
This should be solely decorative and should not affect any functionality.

RazinShaikh:
Should we also support edges between blank nodes? For example, the user might want to highlight a particular part of the diagram by drawing a box around it.

RazinShaikh:
And how should these things work in the proof mode? Can we have different decorations for each step of the proof? There are too many possibilities, we should discuss the desiderata.

jvdwetering:
Yeah it might be good discussing this in a bit more detail in person. I think having different decorations for every step makes a lot of sense. I imagine if you start with some decoration and do a proof step, that this decoration should be copied to the next step. But then there you can modify the decoration while keeping it the same proof step.

PR #380, opened 2025-06-17
Support for blank nodes to add decorations including text and edges

RazinShaikh:
Implements #268
Requires the dummy node implementation from this pyzx PR https://github.com/zxcalc/pyzx/pull/320 

I have added a new node type called dummy which we can use to add textual annotations. We can also use these nodes to draw shapes that don't have semantic meaning but are useful for clarity. 

I have not added support for maths/latex yet. I think it would be better to add latex support in a separate PR.

I would appreciate more feedback on this before we merge it.

### DIFF
--- a/zxlive/base_panel.py
+++ b/zxlive/base_panel.py
@@ -123,6 +123,9 @@ class BasePanel(QWidget):
     def paste_selection(self, graph: GraphT) -> None:
         pass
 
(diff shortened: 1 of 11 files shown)


jvdwetering:
I merged in the PyZX changes so I think this can be merged in now.

PR #387, opened 2025-09-28
Avoid removing eitem twice from self-loop

y-richie-y:
Reproducing the bug:
image

Fuse the spiders with the parallel edges, then drag the Hadamard self-loop. Some of the edges will disappear and we get the following error:
image

```
  File "zxlive/zxlive/graphview.py", line 121, in update_graph
    self.graph_scene.update_graph(g, select_new)
  File "zxlive/zxlive/graphscene.py", line 137, in update_graph
    e_item.t_item.adj_items.remove(e_item)
KeyError: <zxlive.eitem.EItem(0x600003406900, pos=0,0, z=-1, flags=(ItemIsSelectable|ItemSendsGeometryChanges)) at 0x10a44a080>
```

The issue is `GraphScene.update_graph()` removes eitem twice when the source and target are the same:
```python
e_item.s_item.adj_items.remove(e_item)
e_item.t_item.adj_items.remove(e_item)
```

### DIFF
--- a/zxlive/graphscene.py
+++ b/zxlive/graphscene.py
@@ -134,7 +134,8 @@ class GraphScene(QGraphicsScene):
             for anim_e in e_item.active_animations.copy():
                 anim_e.stop()
             e_item.s_item.adj_items.remove(e_item)
-            e_item.t_item.adj_items.remove(e_item)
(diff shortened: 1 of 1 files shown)


RazinShaikh (approved):
lgtm

PR #409, opened 2025-10-17
[WIP] Add fault edge to editor panel

y-richie-y:
My plan is to add support for css fault equivalence analysis by adding an "unidealise edge" action, where the user then has to provide a valid fault equivalent rewrite for that edge. 

Then we can extend this to sets of edges and include detecting region ideas.


RazinShaikh:
Do you first need the pyzx pr to be merged for this? I see that changed the edgetype imports to not import from pyzx

RazinShaikh:
@y-richie-y do you still plan to continue working on this?

PR #431, opened 2026-02-15
Enhance ProofModel by group expansion functionality.

### LINKED ISSUE #292: Expandable group rewrites
Make grouped rewrites like a tree where you can expand. The current implementation is shown below where you can group and ungroup.

axif0:
Enhance ProofStepView with expandable grouped steps and improve item rendering. Added functionality to toggle group expansion on mouse click and updated rendering logic in ProofStepItemDelegate for grouped steps.

Closed: #292 

proof


axif0:
Updated reviews accordingly, thanks for your in detailed suggetion @RazinShaikh .

final

axif0:
Thanks for the suggetion @boldar99 .

zxlive/proof.py
```diff
@@ -6,9 +6,9 @@
 
 from PySide6.QtCore import (QAbstractItemModel, QAbstractListModel,
                             QItemSelection, QModelIndex, QPersistentModelIndex,
-                            QPoint, QPointF, QRect, QSize, Qt)
-from PySide6.QtGui import QColor, QFont, QFontMetrics, QPainter, QPen
-from PySide6.QtWidgets import (QAbstractItemView, QLineEdit, QListView, QMenu,
+                            QPoint, QPointF, QRect, QRectF, QSize, Qt)
+from PySide6.QtGui import QColor, QFont, QFontMetrics, QMouseEvent, QPainter, QPainterPath, QPen, QPolygonF
+from PySide6.QtWidgets import (QAbstractItemView, QInputDialog, QLineEdit, QListView, QMenu,
```
RazinShaikh: some imports are unused

zxlive/rewrite_action.py
```diff
@@ -143,6 +134,28 @@ def do_rewrite(self, panel: ProofPanel) -> None:
         anim_before, anim_after = make_animation(self, panel, g, matches_list, rem_verts_list)
         panel.undo_stack.push(cmd, anim_before=anim_before, anim_after=anim_after)
 
+    # TODO: Narrow down the type of the first return value.
```
boldar99: There is a TODO here

zxlive/rewrite_action.py
RazinShaikh: This PR is unrelated to this file so it shouldn't change anything here.

RazinShaikh (changes requested):
it looks like when you merged main to the current branch, several files (all changes other than proof.py) were merged incorrectly. Can you fix that? Also, can you make sure the checks pass?

boldar99:
Highlighting still seems to block the dots and lines inside the expanded view but not when outside. Could you fix that?

axif0:
>  blue background reaches the very left of the box, not the right of the circle.

Sorry I misinterpreted you earlier. @boldar99 
Updated- 

final

axif0:
Hello, Anything else I need to do on my side?

RazinShaikh (changes requested):

RazinShaikh:
@axif0 thanks for the work so far, can you please refactor such that the lint check passes?

96-LB:
Hi @axif0, sorry for the long delay. Are you still interested in continuing this PR? If so, there is still some work to be done on the code quality. Could you go through and fix the merge conflicts, and make sure the linting checks still pass? Once you've done that, I can give you a code review and work with you to get this ready to merge.

If you don't want to keep working on this since too much time has passed, no worries at all. Just let me know, and I can close the PR.

zxlive/proof.py
```diff
@@ -154,31 +195,47 @@ def get_graph(self, index: int) -> GraphT:
...
+            sub_names = [s.strip() for s in body.split("\u2192")]
+            if len(sub_names) == len(grouped):
+                grouped = [
+                    Rewrite(sub_names[i], rw.rule, rw.graph, rw.grouped_rewrites)
+                    if sub_names[i] != rw.display_name else rw
+                    for i, rw in enumerate(grouped)
+                ]
+        self.steps[index] = Rewrite(name, old.rule, old.graph, grouped)
+        mi = self.createIndex(index + 1, 0, 0)
+        self.dataChanged.emit(mi, mi, [])
```
96-LB, line 212: I don't think it's a good idea to sync sub-step names here. This is highly dependent on the implementation details of the name format, which means it breaks if we change the naming format or the name goes out of sync with the substeps because they were changed after group creation. It can also cause unexpected name changes to substeps for very little gain, since you can just rename substeps directly anyway.

In addition to increasing the complexity of the function, your edits removed the comments that existed here before. Please revert this function to its old state.

zxlive/proof.py
```diff
@@ -154,31 +195,47 @@ def get_graph(self, index: int) -> GraphT:
...
+        self.steps[step_index] = Rewrite(old.display_name, old.rule, old.graph, new_grouped)
+        parent_mi = self.index(step_index + 1, 0)
+        child_mi = self.index(sub_index, 0, parent_mi)
+        self.dataChanged.emit(child_mi, child_mi, [])
 
     def group_steps(self, start_index: int, end_index: int) -> None:
         """Replace the individual steps from `start_index` to `end_index` with a new grouped step"""
         new_rewrite = Rewrite(
-            "Grouped Steps: " + " 🡒 ".join(self.steps[i].display_name for i in range(start_index, end_index + 1)),
+            "Grouped Steps: " + " \u2192 ".join(self.steps[i].display_name for i in range(start_index, end_index + 1)),
```
96-LB, line 231: It would be good to leave a comment here to make it clear what Unicode character this is.

zxlive/proof.py
```diff
@@ -72,57 +73,97 @@ def __init__(self, start_graph: GraphT):
...
+    def columnCount(self, parent: Union[QModelIndex, QPersistentModelIndex] = QModelIndex()) -> int:
+        return 1
 
-        if index.row() >= len(self.steps) + 1 or index.column() >= 1:
-            return None
+    def hasChildren(self, parent: Union[QModelIndex, QPersistentModelIndex] = QModelIndex()) -> bool:
+        return self.rowCount(parent) > 0
 
-        if role == Qt.ItemDataRole.DisplayRole:
+    def _display_name(self, index: Union[QModelIndex, QPersistentModelIndex]) -> Any:
```
96-LB:
```suggestion
    def _display_name(self, index: Union[QModelIndex, QPersistentModelIndex]) -> Optional[str]:
```
This helper function should have its type narrowed. You can keep `ProofModel.data` as is, though.

zxlive/proof.py
```diff
@@ -72,57 +73,97 @@ def __init__(self, start_graph: GraphT):
...
+        if not self.hasIndex(row, column, parent):
+            return QModelIndex()
+        if not parent.isValid():
+            return cast(QModelIndex, self.createIndex(row, column, 0))
+        return cast(QModelIndex, self.createIndex(row, column, parent.row() + 1))
+
+    def parent(self, child: Union[QModelIndex, QPersistentModelIndex]) -> QModelIndex:  # type: ignore[override]
+        if not child.isValid() or child.internalId() == 0:
+            return QModelIndex()
+        return cast(QModelIndex, self.createIndex(child.internalId() - 1, 0, 0))
```
96-LB, line 87: The casts here are unnecessary.

zxlive/proof.py
```diff
@@ -57,11 +58,11 @@ def from_json(json_str: Union[str, Dict[str, Any]]) -> "Rewrite":
...
 
-class ProofModel(QAbstractListModel):
-    """List model capturing the individual steps in a proof.
+class ProofModel(QAbstractItemModel):
+    """Tree model capturing the individual steps in a proof.
 
-    There is a row for each graph in the proof sequence. Furthermore, we store the
-    rewrite that was used to go from one graph to next.
+    There is a top-level row for each graph sequence.
+    Grouped rewrites have child rows representing their individual sub-steps.
```
96-LB, line 65: You removed some information from this docstring. Could you add it back?

zxlive/proof.py
```diff
@@ -72,57 +73,97 @@ def __init__(self, start_graph: GraphT):
...
-        return super().flags(index) | Qt.ItemFlag.ItemIsEditable
+        if not index.isValid():
+            return Qt.ItemFlag.NoItemFlags  # type: ignore[return-value]
+        base = Qt.ItemFlag.ItemIsEnabled | Qt.ItemFlag.ItemIsSelectable
+        if index.internalId() == 0 and index.row() == 0:
+            return base  # type: ignore[return-value]  # START not editable
+        return base | Qt.ItemFlag.ItemIsEditable  # type: ignore[return-value]
 
     def headerData(self, section: int, orientation: Qt.Orientation,
                    role: int = Qt.ItemDataRole.DisplayRole) -> Any:
```
96-LB, line 139:
```suggestion
                   role: int = Qt.ItemDataRole.DisplayRole) -> None:
```

zxlive/proof.py
```diff
@@ -284,27 +377,81 @@ def move_to_step(self, index: int) -> None:
...
+        self.selectionModel().blockSignals(False)
+        step = self.model().steps[step_idx]
+        if step.grouped_rewrites and sub_idx < len(step.grouped_rewrites):
+            graph = step.grouped_rewrites[sub_idx].graph.copy()
+            assert isinstance(graph, GraphT)
+            self.graph_view.set_graph(graph)
+            self.graph_view.graph_scene.selection_changed_custom.emit()
+
+    def proof_step_selected(self, selected: QItemSelection, deselected: QItemSelection) -> None:
+        del deselected  # unused; required by selectionChanged signature
```
96-LB, line 396:
```suggestion
```
We don't `del` unused parameters elsewhere in the codebase, so it's probably best to remove this line for consistency.

zxlive/proof.py
```diff
@@ -284,27 +377,81 @@ def move_to_step(self, index: int) -> None:
         self.graph_view.set_graph(g)
         self.graph_view.graph_scene.selection_changed_custom.emit()
 
+    def navigate_to_sub_step(self, step_idx: int, sub_idx: int) -> None:
```
96-LB, line 380: Rename to `move_to_sub_step` for consistency with `ProofStepView.move_to_step`.

zxlive/proof.py
```diff
@@ -314,44 +461,33 @@ def rename_proof_step(self, new_name: str, index: int) -> None:
                              lambda: self.model().rename_step(index, new_name))
         self.undo_stack.push(cmd)
```
96-LB, line 462: I would swap the order of these two methods, since elsewhere all the substep commands are defined after their counterparts.

zxlive/proof.py
```diff
@@ -511,13 +775,17 @@ def updateEditorGeometry(self, editor: QWidget, option: QStyleOptionViewItem, in
         )
         editor.setGeometry(editor_rect)
 
-    def setEditorData(self, editor: QWidget, index: Union[QModelIndex, QPersistentModelIndex]) -> None:
+    def setEditorData(self, editor: QWidget,
+                      index: Union[QModelIndex, QPersistentModelIndex]) -> None:
         assert isinstance(editor, QLineEdit)
-        value = index.model().data(index, Qt.ItemDataRole.DisplayRole)
-        editor.setText(str(value))
+        editor.setText(str(index.model().data(index, Qt.ItemDataRole.DisplayRole)))
```
96-LB, line 781: No reason to touch this method in this PR.

96-LB (commented):
Thanks for the update @axif0. I've gone through and done an initial pass on the code and left a review. Overall, it's looking much improved!

I also have a few global comments:
- Normally, the graph displayed when you click on the group rewrite matches the graph of the last child, and this is enforced in a few spots in the code. However, if you edit the graph in-place (i.e., drag the vertices around) of either rewrite, these two displays become desynchronised. This can cause some rather confusing behaviour, and it would be best to make sure these two graphs are always the same when you call `set_graph` or `set_sub_graph`.
- In many places, you use `# type: ignore` statements that are not necessary. I believe all of them except the override can be removed without impact. Since these suppress linting errors it's best to only use them when absolutely necessary.
- Your PR adjusts the formatting of several methods which aren't modified. Please revert these changes to keep your PR as focused as possible.
- Can you add docstrings for the methods you added which are still missing them?
- Currently, it is possible to create nested groups, but the UI doesn't support this and the original rewrites become inaccessible. ~~Since the UI is a bit confusing in this case, I think it would be best to avoid allowing grouped commits to be further grouped. However, nested groups were also possible before this PR -- @RazinShaikh, do you agree that this makes sense to remove?~~ (Edit: see Razin's comment below!)

Image

RazinShaikh:
> Currently, it is possible to create nested groups, but the UI doesn't support this and the original rewrites become inaccessible. Since the UI is a bit confusing in this case, I think it would be best to avoid allowing grouped commits to be further grouped. However, nested groups were also possible before this PR -- @RazinShaikh, do you agree that this makes sense to remove?

I like the ability to group the grouped steps so I am not keen on removing them. If that's not allowed, then one would need to ungroup grouped steps and then regroup into a larger group. There are two solutions we can consider here: 
1. Allow nested group in the UI so the interface looks like a tree similar to what a file explorer view does:
image

2. When grouping a grouped rewrite, flatten the group: [[r1,r2], r3] becomes [r1, r2, r3]

My preference is for the first option where the UI supports nested rewrites but if you want to go with the second option, that's fine too as it's easier to implement.

Issue #388, opened 2025-09-29
Support LaTeX expressions on blank nodes


RazinShaikh:
It should build on top of https://github.com/zxcalc/zxlive/pull/380

jvdwetering:
Do you mean you want to render the correct thing inside the gui? We could do the half solution that tikzit uses where certain things get replace by a unicode counterpart (like greek letters)

RazinShaikh:
We could use some existing package to convert latex to svg and then display the svg inside the gui

RazinShaikh:
@axif0 Congrats on winning this bounty 🥳

PR #432, opened 2026-02-15
Support LaTeX expressions on blank nodes

axif0:
Adds LaTeX rendering support for dummy (blank) node labels. When a user double-clicks a dummy node and enters LaTeX markup (e.g. `\alpha`, `\ket{0}`, `x^2`, `\frac{1}{2})`, the text is detected as LaTeX and rendered as unicode with HTML superscripts/subscripts via pylatexenc — a pure-Python library covering thousands of LaTeX commands including Greek letters, math operators, accents, `\mathbb`, and Dirac notation (`\ket`, `\bra`, `\braket`). 

Plain text labels continue to work unchanged. Rendering results are cached per-node to avoid redundant conversion on repeated `refresh()` calls, and dark mode is supported through color-aware HTML output. The new [pylatexenc](https://github.com/phfaist/pylatexenc) dependency is added as a required package (zero transitive dependencies, MIT licensed).

closing #388

### DIFF
--- a/test/test_latex_render.py
+++ b/test/test_latex_render.py
@@ -1,0 +1,203 @@
+"""Unit tests for zxlive.latex_render module.
+
+Covers is_latex detection, Dirac notation preprocessing (including the
+two-argument \\braket{a}{b} form), SVG rendering, and fallback behaviour.
+"""
+
+from zxlive.latex_render import (
+    is_latex,
+    _preprocess_dirac,
+    _fallback_svg,
+    _svg_viewbox_empty,
+    latex_to_svg,
+)
+
+
+# Tests to verify if a given string is correctly identified as LaTeX markup
+
+class TestIsLatex:
+    """Tests for the is_latex() detection function."""
+
+    def test_empty_string(self) -> None:
+        assert not is_latex("")
+
+    def test_plain_text(self) -> None:
+        assert not is_latex("hello world")
+
+    def test_dollar_sign(self) -> None:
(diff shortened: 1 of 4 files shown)


axif0:
We can use the feature like -

latex

axif0:
Sorry, I thought we'll use pure python for the Codebase. Now we'll use KaTex.

image

KaTeX needs Node.js and the `katex`  npm package. We need to install `katex` by `npm install katex`. KaTeX outputs MathML, which is then converted to SVG `<text>/<line>` elements in pure Python, compatible with Qt's SVG Tiny renderer

RazinShaikh:
Ah I see! Sorry, I was not aware that KaTeX outputs mathml instead of svg. In that case, adding a over 500 lines worth of mathml to svg conversion code (probably ai generated) is not a good idea. If you don't mind, we should roll back to your earlier implementation which uses Ziamath. Sorry for the trouble.

zxlive/dialogs.py
```diff
@@ -227,7 +227,7 @@ def get_file_path_and_format(parent: QWidget, filter: str, default_input: str =
             show_error_msg("Unable to determine file format.", parent=parent)
             return None
 
-    # Add file extension if it's not already there
+    # Add file extension path if missing
```
RazinShaikh: can you undo this comment, the original is easier to understand in my opinion

RazinShaikh:
@axif0 can you comment on the issue so that we can assign it to you ?

Issue #425, opened 2026-01-06
Proof step thumbnails like a filmstrip view


RazinShaikh:
When working on a proof in ZXLive, we can only see one ZX diagram at a time. This is unlike working on a whiteboard or a notebook or tikzit where we can see multiple steps of the proof. Having this global view is really important to maintain the chain of thought while doing minor rewrites at a given time. ZXLive allows the user to switch through proof steps but we can still only see one diagram at once. 

We should make it easier for the user to look at multiple steps at once. One way to achieve this is to have a global proof view where we can see a thumbnail/preview of a diagram at each step. This should be part of the proof step view, which currently only shows the name of the rewrite. There should be an option to toggle diagram preview on and off. It is important that it is easy to resize the size of the preview diagrams.

<img width="2837" height="1565" alt="Image" src="https://github.com/user-attachments/assets/43d1d81f-e8c2-4471-a2ba-488e39b3fdfb" />

A potential starting point to be convert the diagram to a thumbnail is to look at the following code which generates previews for custom rule tooltips.

https://github.com/zxcalc/zxlive/blob/ac5f541c9995a7e64580d240268a5ab0ba748185/zxlive/rewrite_action.py#L191-L226

PR #434, opened 2026-02-16
Add thumbnail preview functionality to proof panel and model

axif0:
- Introduced a toggle button in the ProofPanel for displaying thumbnails of proof steps.
- Implemented thumbnail rendering logic in the ProofModel, including caching for performance.
- Updated settings to manage thumbnail visibility and added related methods in DisplaySettings.
- Adjusted ProofStepView and ProofStepItemDelegate to accommodate thumbnail display.

Closes: #425 

thumb

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,11 +37,12 @@ dependencies = [
     "imageio",
     "packaging",
     "matplotlib>=3.3",
+    "galois",
 ]
 
 [project.optional-dependencies]
 test = [
-    "mypy",
+    "mypy==1.19.1",
     "ruff",
     "pytest-qt",
     "complexipy",
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pyzx>=0.8.0
+pyzx>=0.10.0
 networkx~=3.1
 numpy~=1.25.2
 pytest-qt~=4.2.0
@@ -6,7 +6,7 @@ PySide6~=6.7.2
 shapely~=2.0.1
 shapely-stubs @ git+https://github.com/ciscorn/shapely-stubs.git
 pyperclip>=1.8.1
-mypy
+mypy==1.19.1
 pylint
 pyproject-flake8
 sphinx>=2.3,<7  # see: https://github.com/readthedocs/readthedocs.org/issues/10279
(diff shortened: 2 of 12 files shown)


boldar99:
> > I'm not able to run the most recent commit of this PR. I get the error:
> > >
> 
> I updated it using `typing.Union` . I think it is a OS issue. 😅

The `|` notation for typing was introduced in Python 3.10 (https://peps.python.org/pep-0604/), and zxlive supports python versions >=3.9, so that's probably the reason Lia was not able to run this.

axif0:
> you know if that's too resource intensive and might make ZXLive slow?

Ya, I feel the app a bit lag later ~

When you drag the proof panel's splitter, `resizeEvent` fires hundreds of times per second, each one would re-render all thumbnails from scratch.

Instead, every `resizeEvent` now just restarts a 150ms one-shot timer. Only when you stop dragging does the timer actually fire, triggering a single cache-clear and re-render at the final size. 
Result: zero churn during the drag, one crisp re-render at the end.

Is it a good way ? @RazinShaikh

axif0:
Sorry, for the cause, `drawPixmap(x, y, pixmap)` draws at raw physical-pixel size, ignoring DPR. Different screens = different sizes.

image

axif0:
Slightly bumped the max render resolution and now pass the raw high-res pixmap directly into `painter.drawPixmap()` . This lets Qt natively scale the image using all available physical pixels for Retina displays, eliminating the need for any of that complex, buggy resize logic. which solves the issue i think.

RazinShaikh:
@axif0  please address copilot's review comments

axif0:
Is there anything else I can do? 😅 @RazinShaikh

RazinShaikh:
The thumbnails are still not rendering properly. The following screenshot shows that the thumbnails do not display the whole diagram.

image

RazinShaikh:
@axif0  Can you merge master into this branch? I think that should fix the tests.

Also, it's taking me long time to merge this because since this is a big PR and has a lot of changes, it's difficult to get the code quality to where it needs to be. In particular, I am not happy with the changes in proof.py as some of the handling feels a bit hacky. I can work on improving it but it will take me some time to get to it as I am a bit occupied currently. You are also welcome to try to improve it but I suggest not using AI tools for that as that will just continue to make the code quality worse.

PR #435, opened 2026-02-17
Add diff highlighting for proof step differences in the graph view

Kitsunp:
Addresses #190
Adds visual highlighting of vertices and edges that will change between
proof steps. When navigating to a step in proof mode, elements affected
by the next rewrite are highlighted with coloured outlines and tinted
fills, making it easy to spot the difference without manual comparison.

**What it does:**
- Vertices that will be removed get a crimson outline and tinted fill
- Vertices whose type or phase will change get an orange outline and tinted fill
- Edges that will be removed or change type are highlighted similarly
- Only semantically meaningful changes are shown (type, phase, structural);
  cosmetic changes like repositioning are excluded to avoid visual noise

**How to test:**
1. Open a proof, apply a few rewrites
2. Click on any step that is not the last one
3. The nodes/edges that change in the next step should be visibly highlighted
4. Clicking the last step clears all highlights

**Files changed:**
- `settings.py` — diff colour scheme + toggle setting
- `settings_dialog.py` — checkbox for "Highlight proof step differences"
- `vitem.py` — diff highlight rendering for vertices
- `eitem.py` — diff highlight rendering for edges
- `graphscene.py` — `highlight_diff()` and `clear_diff_highlights()` methods
- `proof.py` — calls highlighting in `move_to_step()`

Leverages the existing `pyzx.graph.diff.GraphDiff` already used in
`GraphScene.update_graph()`. No new dependencies. Toggleable via Settings.


zxlive/rewrite_data.py
RazinShaikh: all the changes in this file are not related to the diff highlighting for proof steps

PR #445, opened 2026-02-22
Highlight difference in rewrite steps

kwyip:
Aim to fix https://github.com/zxcalc/zxlive/issues/190

1. Introduced apply_step_difference_highlighting(current, next_graph) to compute and apply diff_highlight flags on the currently displayed graph. 
2. Vertices: when vdata(..., "diff_highlight") is true, draw thicker yellow outlines (selected and unselected states handled). 
3. Edges: when edata(..., "diff_highlight") is true, draw a yellow overlay stroke. 

For example, in fusing two nodes the new node and a neighboring node are highlighted, as in Lia's [example](https://github.com/zxcalc/zxlive/issues/190#issuecomment-3715930499).

But...

- Position-only changes do not highlight. 
- Equivalent geometry with different vertex IDs does not highlight. 

https://github.com/user-attachments/assets/830ddce6-4ecb-42d6-9ffc-a1819066091f

### DIFF
--- a/zxlive/proof.py
+++ b/zxlive/proof.py
@@ -12,7 +12,7 @@ from PySide6.QtGui import QAction, QColor, QFont, QFontMetrics, QPainter, QPen
 from PySide6.QtWidgets import (QAbstractItemView, QLineEdit, QListView, QMenu,
                                QStyle, QStyledItemDelegate,
                                QStyleOptionViewItem, QWidget)
-from .common import GraphT
+from .common import GraphT, get_settings_value
 from .settings import display_setting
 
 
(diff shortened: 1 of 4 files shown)


lia-approves:
Hello, thanks for the PR. It has the issue that Boldi described in #435 :
> In particular, when two spiders are fused, it is the nodes and the connecting edge that should be highlighted.

Nodes that don't participate in teh rewrite rule should not be highlighted. Let us know if you have any questions.

boldar99:
Other example with fusion:


https://github.com/user-attachments/assets/b82a3a06-0e27-4af8-9c51-1e6d15f92d63

PR #479, opened 2026-03-29
fix stale phase labels on boundary vertices

dlyongemallo:
Relates to #462.


dlyongemallo:
CI failure is due to `lint / code_complexity` and is fixed by #490.

dlyongemallo:
> Copilot's observation is correct. 

We are truly living in the future.

PR #485, opened 2026-03-31
add toggleable vertex index labels

### LINKED ISSUE #342: Toggleable vertex labels
Suggested feature: a toggle under the 'view' tab (or perhaps in 'preferences') to enable vertex index labels (as in pyzx)

dlyongemallo:
Fixes #342.

### DIFF
--- a/zxlive/vitem.py
+++ b/zxlive/vitem.py
@@ -20,7 +20,7 @@ import math
 from typing import Optional, Set, Any, TYPE_CHECKING, Union
 
 from PySide6.QtCore import Qt, QPointF, QVariantAnimation, QAbstractAnimation, QRectF
-from PySide6.QtGui import QPen, QBrush, QPainter, QColor, QPainterPath
+from PySide6.QtGui import QPen, QBrush, QPainter, QColor, QFont, QPainterPath
 from PySide6.QtSvg import QSvgRenderer
 from PySide6.QtSvgWidgets import QGraphicsSvgItem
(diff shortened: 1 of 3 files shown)


dlyongemallo:
As an example of what this looks like:

image

zxlive/vitem.py
```diff
@@ -215,7 +217,41 @@ def _refresh_w_partner(self, ty: VertexType, cascade_edges: bool) -> None:
...
+            angle = math.radians(self.rotation())
+            cos_a = math.cos(angle)
+            sin_a = math.sin(angle)
+            offset = 0.25 * SCALE
+            rect = self.index_text_item.boundingRect()
+            shift_x = -rect.width() / 2
+            self.index_text_item.setPos(
+                shift_x * cos_a + offset * sin_a,
+                -shift_x * sin_a + offset * cos_a)
+            self.index_text_item.setVisible(True)
```
RazinShaikh: I don't understand this calculation. Why is cos_a and sin_a necessary for positioning the label below the vertex?

dlyongemallo: It's because the labels need to be transformed into their parents' local coordinates, which may themselves be rotated, so the labels need to be rotated back. Line 229 counter-rotates the label's position, and lines 236-238 counter-rotate the rotation. 

Although, looking at `set_vitem_rotation`, the only possible vertex type to have a rotation is `W_OUTPUT`, so I'll gate the counter-rotation logic based on that.

RazinShaikh: Sounds good, thanks

PR #487, opened 2026-04-08
Make saving rewrite in rules folder the default

### LINKED ISSUE #486: Open file explorer in the rewrite folder when saving a rewrite rule
When saving a rewrite rule, the file explorer should be start in the rewrite folder (available from the settings), making it easier for the user simply specify a name and hit save.

satkarjuneja:
Fixes #486 
When saving a rewrite rule, the file explorer now opens directly in the custom rules folder (configured in Settings → General → Custom rules path) instead of the system default directory.

The setting and the parameter to pass it through already existed : ` _save_rule_or_proof_dialog` already accepted a dir argument, and path/custom-rules was already in settings. This just wires them together in save_rule_dialog.

### DIFF
--- a/zxlive/dialogs.py
+++ b/zxlive/dialogs.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+import os
 from dataclasses import dataclass
 from enum import Enum
 from typing import TYPE_CHECKING, Optional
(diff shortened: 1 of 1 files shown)


RazinShaikh:
I believe this is for #486 not #458

RazinShaikh:
I'll edit the description

satkarjuneja:
> I believe this is for #486 not #458

Oops!, Sorry I was looking at that one before hand and accidentally copied that one

PR #488, opened 2026-04-08
generalise `get_embedded_app` for use outside Jupyter

dlyongemallo:


dlyongemallo:
@jvdwetering @RazinShaikh Do you have or foresee any need to run ZXLive from Python outside of a Jupyter notebook? 

I made this draft PR for the [TQEC/topologiq](https://github.com/tqec/topologiq) developers to experiment with, as an illustration. It's not exactly what they need, as what they want is much tighter integration. (They want something embedded into their UI or workflow to edit ZX diagrams, rather than what this does, which is run ZXLive as a modal popup, which interrupts their workflow, and when closed resumes it.) 

However, if there is a need to run ZXLive from Python anywhere, I could test this some more and polish it up and make an actual PR out of it. Alternatively or in addition, maybe a feature could be added to ZXLive to allow what the topologiq developers want (fully embed ZXLive inside another app), though that's a much larger undertaking requiring API design changes.

jvdwetering:
I personally don't see any use-case for running this outside of a jupyter notebook. But who knows, maybe there is someone who just wants to run it from a regular python terminal. If you have already done the work to make this possible I don't see the harm in why this shouldn't be adopted.

I'm not sure what it would take to run it 'fully embedded', but that does indeed sound like a much larger project.

Issue #303, opened 2024-07-09
Add animation for hopf rule


RazinShaikh:
We can apply hopf rule, aka leg chopping, using the magic wand but this rewrite is missing an animation.

jvdwetering:
What should the animation look like? The wires gradually getting thinner?

RazinShaikh:
Doing something like the following would be very cool.

<img width=250 src="https://github.com/zxcalc/zxlive/assets/28257517/7bc56bd2-7107-4b33-a8cc-39a97bd9a915" /><br>
<img width=250 src="https://github.com/zxcalc/zxlive/assets/28257517/f69a3d35-5b7e-4819-8b0e-2a862d67ba19" /><br>
<img width=250 src="https://github.com/zxcalc/zxlive/assets/28257517/d07b3e11-ce56-4385-90e0-d7d2b6c45aab" /><br>
<img width=250 src="https://github.com/zxcalc/zxlive/assets/28257517/292c9051-c40c-41d5-9ff4-384ab60cad70" />

jvdwetering:
That would make me feel like a ninja, so I approve. That seems complicated though. Maybe something: calculate intersection point with the slice, then replace each of the edges with two edges where the boundary point is at the slice point. Then move this point closer to the spider until you can remove it. However, the hard part would be figuring out how to curve the wires properly as they move back.

dlyongemallo:
I implemented a simple fade-out in #496. I also made a more involved version with the ninja-like slashing, but it was like another 180 lines of code so I didn't think it was worth it. But if you _really_ want the ninja effect, I can change it to that version.

jvdwetering:
I guess it depends: how cool did the ninja effect look? :)
Making rewriting *fun* is an explicit goal of ZXLive, so if it does make you feel like a ninja and you've already wrote it we should consider it.

RazinShaikh:
I don't mind the simple fade-out :)

PR #496, opened 2026-04-27
add simple fade animation for hopf rule (parallel edge removal)

dlyongemallo:
Relates to #303.

### DIFF
--- a/zxlive/animations.py
+++ b/zxlive/animations.py
@@ -2,11 +2,13 @@ from __future__ import annotations
 
 import itertools
 import random
+from collections import Counter
 from typing import Optional, Callable, TYPE_CHECKING
 
 from PySide6.QtCore import QEasingCurve, QPointF, QAbstractAnimation, \
     QParallelAnimationGroup
 from PySide6.QtGui import QUndoStack, QUndoCommand
+from pyzx.graph.diff import GraphDiff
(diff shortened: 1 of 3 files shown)


RazinShaikh:
This animation seem to only apply when using the magic wand but not when using the rule "Remove parallel edges" from the sidebar. Can you please link the animation to that rule as well?

RazinShaikh:
I made a slight modification by adding the opacity and reducing it to 0 in the animation along with reducing the edge width. This makes the animation look more natural. I hope you don't mind. Also, since the opacity is made 0 at the end of the animation, it is no longer necessary to hide the edges.

dlyongemallo:
But you still need to this be hooked up to the rule from the sidebar, right? Or did you make/will you be making that change?

RazinShaikh:
I think you already did that, didn't you? After your last commit, the "remove parallel edges" from the sidebar works. I only modified the style of the fade.

dlyongemallo:
Oh, right, the last push already included the sidebar fix. I had this MR marked as not completed in my list and got confused when I looked over the comments, but the not-yet-done thing was to evaluate the slash/ninja animations mentioned in #303. I think the fade-out is fine, but I can also follow up with a PR for fancy slashing when I have time (low priority), if you want to evaluate whether it's worth the additional complexity.

PR #500, opened 2026-04-27
add auto-arrange feature using `networkx` spring layout

### LINKED ISSUE #326: Command to auto-arrange nodes in the diagram
This can be done by calling vertex arrangement function from some library like networkx

dlyongemallo:
Fixes #326.

### DIFF
--- a/test/test_mainwindow.py
+++ b/test/test_mainwindow.py
@@ -20,8 +20,10 @@ from pathlib import Path
 from PySide6 import QtCore
 from pytestqt.qtbot import QtBot
 
+from pyzx.utils import EdgeType, VertexType
+
 from zxlive.dialogs import import_diagram_from_file
-from zxlive.common import new_graph
+from zxlive.common import GraphT, W_INPUT_OFFSET, new_graph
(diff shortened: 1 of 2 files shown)


jvdwetering:
Nice work! A couple of questions: what happens to nodes not connected to anything? In any case would be good to have a test case covering that. And what happens if you have multiple boundary nodes on top of each other?
Would it be possible to make this selection-aware, so that you only reposition whatever you have selected, and you treat the outgoing connections to neighbouring vertices of what you have selected as the boundary vertices.

dlyongemallo:
> what happens to nodes not connected to anything? In any case would be good to have a test case covering that. 

Disconnected nodes stay near their origin location. Added a test case for this.

> And what happens if you have multiple boundary nodes on top of each other?

Multiple overlapping boundary vertices stay where they are.

> Would it be possible to make this selection-aware, so that you only reposition whatever you have selected, and you treat the outgoing connections to neighbouring vertices of what you have selected as the boundary vertices.

Yes, done: this is now selection-aware.

PR #501, opened 2026-04-27
deduplicate `requirements.txt` by referencing `pyproject.toml` extras

### LINKED ISSUE #163: differentiate between deployment and development requirements
Right now, there are two sets of requirements, one in `requirements.txt` and one in `pyproject.toml`. I think probably the former should be deprecated in favour of the latter, which is more expressive. Either way, they should be merged.

The requirements as currently stated are *development* requirements (as they include required packages for testing, linting, and typing). For creating an executable (see #158), it's necessary to identify the minimum set of *deployment* requirements, to make the resulting executable as small as possible. There's obviously no reason to include typing stubs or testing frameworks into the deployed executable.

(Also, I noticed that `sympy` is stated as a requirement but isn't used anywhere. Is it intended to be used in the future, or can it be removed from the requirements?)

dlyongemallo:
Closes #163.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,17 +1,5 @@
-pyzx>=0.10.0
-networkx~=3.1
-numpy~=1.25.2
-pytest-qt~=4.2.0
-PySide6~=6.7.2
-shapely~=2.0.1
-shapely-stubs @ git+https://github.com/ciscorn/shapely-stubs.git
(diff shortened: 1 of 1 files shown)


RazinShaikh:
Is it okay if we simply delete the requirements.txt?

dlyongemallo:
I have no objection to that.

PR #506, opened 2026-04-30
preserve selection across `ProofModeCommand` operations

### LINKED ISSUE #505: in proof mode, operations cause the selection to be lost
I noticed this while using the new auto-arrange functionality, but even just dragging a selected vertex or group of selected vertices causes the selection to be lost.

dlyongemallo:
Fixes #505.

### DIFF
--- a/test/test_mainwindow.py
+++ b/test/test_mainwindow.py
@@ -547,6 +547,184 @@ def test_auto_arrange_in_proof_mode(app: MainWindow, qtbot: QtBot) -> None:
     assert (proof_graph.row(spider_v), proof_graph.qubit(spider_v)) == spider_after
 
 
+def _new_two_spider_proof_panel(app: MainWindow, qtbot: QtBot) -> tuple[ProofPanel, int, int]:
+    # Open a proof panel on a small graph with two internal Z spiders and return their ids.
+    app.close_action.trigger()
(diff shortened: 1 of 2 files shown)


RazinShaikh:
There is a slight inconsistency with this implementation. For example, if we use rules like spider fusion, then the selection is appropriately preserved. But for some other rules where the rewrite adds new nodes, the behaviour is less clear. For example, the push pauli rule will only keep the non-pauli spider selected because the old pauli spider which was copied to create two new pauli spiders:

image 
image

On the other hand, strong complementarity will unselect all the nodes because in the implementation, new spiders are created and the code does not know which new spiders are spawned from the older selected ones.

Do you think it's possible to do something better? If not, I am also fine with just keeping what you have currently.

RazinShaikh:
I also noticed another bug. If we preserve the selection, the sidebar is not updated with the selection and it shows the rules that were applicable from the previous selection and it might allow us to apply a rule incorrectly.

dlyongemallo:
@RazinShaikh Can you give this another try?

RazinShaikh:
Hi, sorry I only just got a chance to test it. It works now without the issues I mentioned above.

RazinShaikh (approved):

PR #521, opened 2026-05-13
Edge tool can draw multiple parallel edges at once

RazinShaikh:


jvdwetering:
What is the use-case of being able to draw multiple parallel edges in one go?

RazinShaikh:
There was a feature request in the ZX discord about this feature a couple of months ago but I finally got a chance to clean up the implementation and merge it.

https://github.com/user-attachments/assets/bbb24e7d-fac0-4db4-ac15-aaa0c8f2fa4b


One use case is this: Imagine you have a 20 qubit circuit with many gates. You can now select the input 20 boundary nodes on the input and connect them to the 20 output boundary nodes in one go. Previously, you would have to repeat this process 20 times. 

Another use case: you can draw all the edges for transversal cnots or a cnot ladder in one go.

jvdwetering:
ooohh, I misunderstood the feature description. Yes, agreed this is very useful.

PR #530, opened 2026-06-03
feat: add interactive tutorial overlay to guide new users through the…

### LINKED ISSUE #523: Add an interactive onboarding tutorial
Add an interactive tutorial that overlays on top of the UI and guides users through the main features of the app.

The tutorial should explain the purpose of key UI components and tools, such as:

- What tools like Add Vertex and Add Edge do
- What each sidebar is used for
- How to double-click a vertex to add a phase
- How to enter and use proof mode
- How to apply rewrites
- What the magic wand tool does
- etc...

Users should be able to click Next (or Previous) to move through each step of the tutorial. The goal is to help new users understand everything they need to get started without overwhelming them with too much information at once.

The tutorial should also include:

- A Skip button for users who do not want to complete it
- An option to replay the tutorial from the Help menu
- A proof-mode-specific section that appears when the user first enters proof mode

In order to prevent bloating the code, as much as possible the tutorial should consist of a separate source file leading the user through the tutorial, instead of adding a little bit of functionality to many different files for enabling this tutorial.

Gmin2:
adds an interactive tutorial that overlays the UI and walks new users through the main features. it dims the screen, spotlights the real toolbar buttons, sidebars and vertices, and explains each one with a small coachmark you move through with next/back/skip. the build steps (add a vertex, draw an edge, set a phase) wait for you to actually do the action before advancing, with a ghost pointer showing where to click. it runs on a clean canvas, auto-starts on first launch, and can be replayed any time from Help > Tutorial. there's also a proof-mode section that fires the first time you start a derivation, covering the magic wand, rewrites and proof steps.

almost all of it lives in one new file, zxlive/tutorial.py, so the panels stay untouched. the only changes elsewhere are a Help menu entry, the first-run trigger in app.py, and one line in mainwindow where proof mode opens. seen-state is stored in QSettings so it doesn't nag returning users.

https://github.com/user-attachments/assets/1af16c62-e056-46cf-af39-e012e8a30546

Closes #523

### DIFF
--- a/zxlive/tutorial.py
+++ b/zxlive/tutorial.py
@@ -18,29 +18,36 @@
 Overlays the existing UI instead of touching the panels: a dimming spotlight
 punches click-through holes around the real toolbar buttons, sidebars and
 vertices, and a coachmark walks through them. On the build steps it waits for
-the user to actually do the action before advancing.
+the user to actually do the action before advancing (revisiting a completed
+step via Back does not require redoing it).
 
-The rest of the app only calls start_main_tutorial (Help menu),
-maybe_show_tutorial_on_first_run (first launch) and maybe_start_proof_tutorial
-(first time proof mode opens).
+MainWindow owns a TutorialController instance; the rest of the app only calls
+start_main_tutorial (Help menu), maybe_show_tutorial_on_first_run (first
+launch) and maybe_start_proof_tutorial (first time proof mode opens).
 """
 
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Callable, Optional, TYPE_CHECKING
+from typing import Callable, Optional, Sequence, TYPE_CHECKING, Union
 
 from PySide6.QtCore import (QEasingCurve, QEvent, QObject, QPoint, QPropertyAnimation,
-                            QRect, QRectF, Qt, QTimer, QVariantAnimation)
-from PySide6.QtGui import QColor, QFont, QPainter, QPainterPath, QPen, QPolygon
-from PySide6.QtWidgets import (QFrame, QGraphicsDropShadowEffect, QHBoxLayout, QLabel,
-                               QPushButton, QVBoxLayout, QWidget)
+                            QRect, QRectF, Qt, QTimer, QVariantAnimation, SignalInstance)
+from PySide6.QtGui import (QColor, QFont, QMouseEvent, QPaintEvent, QPainter, QPen,
+                           QRegion)
+from PySide6.QtWidgets import (QFrame, QGraphicsDropShadowEffect, QGraphicsItem,
+                               QHBoxLayout, QLabel, QPushButton, QToolButton,
+                               QVBoxLayout, QWidget)
+from pyzx.utils import VertexType
 
-from .common import get_settings_value, set_settings_value
+from .common import GraphT, get_settings_value, new_graph, set_settings_value
+from .edit_panel import GraphEditPanel
+from .proof_panel import ProofPanel
 from .settings import display_setting
 
 if TYPE_CHECKING:
     from .base_panel import BasePanel
+    from .graphview import GraphView
     from .mainwindow import MainWindow
 
 
@@ -53,6 +60,10 @@ ACCENT = "#8a63ff"
 _HOLE_PADDING = 8
 _CARD_WIDTH = 340
 
+# A step can spotlight widgets and/or graphics items (e.g. a VItem in the scene).
+TargetItem = Union[QWidget, QGraphicsItem]
+TargetSpec = Union[TargetItem, Sequence[Optional[TargetItem]], None]
+
 
 # Step definition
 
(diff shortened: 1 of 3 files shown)


zxlive/tutorial.py
```diff
@@ -0,0 +1,782 @@
...
+        self._disconnect_await()
+        self.coachmark.flash_done()
+        self._celebrate()
+        QTimer.singleShot(750, self._auto_advance)
+
+    def _auto_advance(self) -> None:
+        # Only advance if we're still on the same (now-completed) step.
+        if self.active and not self._awaiting:
+            self._next()
+
```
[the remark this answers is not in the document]
96-LB: I was able to reproduce this issue. It can happen if an "awaiting" step is followed by a "non-awaiting" step, or if you manage to complete the action of the next step within the 750ms.

zxlive/tutorial.py
```diff
@@ -0,0 +1,782 @@
...
+
+# Entry points (the only things the rest of the app calls)
+
+def _controller(win: "MainWindow") -> TutorialController:
+    controller = getattr(win, "_tutorial_controller", None)
+    if controller is None:
+        controller = TutorialController(win)
+        win._tutorial_controller = controller  # type: ignore[attr-defined]
+    return controller
```
96-LB: Instead of injecting an extra attribute onto the object here, it would be nice if the MainWindow class itself defined and instantiated the tutorial controller so that type-checkers can see the attribute.

zxlive/tutorial.py
```diff
@@ -0,0 +1,782 @@
...
+def _first_vertex(c: TutorialController) -> object:
+    panel = c.panel()
+    if panel is None:
+        return None
+    vmap = panel.graph_scene.vertex_map
+    return next(iter(vmap.values()), None) or panel.graph_view
+
+
+def _attr(c: TutorialController, name: str) -> object:
+    return getattr(c.panel(), name, None)
```
96-LB: This function potentially hides a lot of errors—using `getattr` prevents the type-checker from validating attribute references, and casting everything as `object` prevents the rest of your code from having useful type annotations (you write `object` everywhere even though the code is only concerned with QWidgets). Even if it makes the definition of the step list a little longer, I think it's much more worth it to explicitly write out all the attribute references and properly type-hint the code.

zxlive/tutorial.py
```diff
@@ -0,0 +1,782 @@
...
+        if self._awaiting:
+            try:
+                signal = step.await_event(self)
+            except Exception:
+                signal = None
+            if signal is not None:
+                signal.connect(self._on_await)
+                self._await_obj = signal
+            else:
+                self._awaiting = False
```
96-LB:
```suggestion
        self._awaiting = False
        if step.await_event is not None:
            try:
                self._await_obj = step.await_event(self)
                self._await_obj.connect(self._on_await)
                self._awaiting = True
            except AttributeError:
                self._await_obj = None
```

Putting the condition in the if statement allows the type-checker to infer that `step.await_event` is not `None`. 

There's still a type error here because `self._await_obj` is typed as object; see my other comment.

zxlive/tutorial.py
```diff
@@ -0,0 +1,782 @@
...
+            title="Welcome to ZXLive 👋",
+            text="ZXLive lets you <b>draw</b> ZX-diagrams and then <b>prove</b> "
+                 "things about them. This quick tour takes about a minute — you "
+                 "can hit <i>Skip</i> any time.",
+        ),
+        Step(
+            title="The canvas",
+            text="This is your diagram. Drag the background to pan, "
+                 "<b>Ctrl + scroll</b> to zoom, and press <b>C</b> to fit "
+                 "everything in view.",
```
96-LB:
```suggestion
            text="This is your diagram. <b>Scroll</b> to pan, "
                 "<b>Ctrl + scroll</b> to zoom, and press <b>C</b> to fit "
                 "everything in view.",
```

Dragging the background is used for multi-selecting objects (unless this varies by OS?).

zxlive/tutorial.py
```diff
@@ -0,0 +1,782 @@
...
+            title="Proof steps",
+            text="Each rewrite you apply shows up here. Click any step to jump "
+                 "back and forth through your proof — nothing is ever lost.",
+            target=lambda c: _attr(c, "step_view"),
+        ),
+        Step(
+            title="Rewrites",
+            text="These are the rules you can apply: fuse spiders, change "
+                 "colours, remove identities and more. Hover a rule to preview "
+                 "it, and right-click to apply it to your selection.",
```
96-LB:
```suggestion
            text="These are the rules you can apply: fuse spiders, change "
                 "colours, remove identities and more. Hover a rule to preview "
                 "it, and click to apply it to your selection.",
```

96-LB (commented):
Thanks for the PR! The design looks very nice. I've done a preliminary review and left some specific comments. Also make sure to consider the comments left by Copilot.

Some general feedback:
- Perhaps it would be useful to start the tutorial with a basic graph on the canvas. The first step feels a little confusing since there is no diagram to press "C" to center on.
- Sometimes you refer to keys in lowercase and sometimes in uppercase.
- During the tour, the panels did not light up properly on my machine until they were clicked on. You may need to find a way to force Qt to re-render your spotlight graphics.
- When pressing "Back" to view previous steps, you have to perform all the "Your turn!" actions a second time to proceed. It would be nice to avoid making the user repeat the interactive steps.
- In general, it would be helpful if your code is properly type-annotated. There are a lot of places that your variables are annotated as `object` or you use `#type: ignore`, which makes the code difficult to maintain because the type-checker cannot reason about it and errors get hidden.

Gmin2:
Thanks for the review! addressed everything in 0627d92, the race fix, animation reuse, proper type annotations, starter graph on the canvas and the copy fixes. back steps also dont need redoing anymore.

cc @96-LB

Gmin2:
cc @96-LB

96-LB:
Thanks for your contribution! We will get this PR merged shortly.

Could you send a message in the thread for the issue #523 and ping me in it? Once you do so, I can assign you to the issue so that you can be awarded the unitaryHACK bounty.

Gmin2:
hey any update ?

PR #535, opened 2026-06-05
Add interactive onboarding tutorial

harsh-791:
Addresses #523.

## Summary

Adds an interactive onboarding tutorial: a guided overlay that dims the window,
spotlights one UI element at a time (with a gentle pulse and a card that points
at the target), and shows an explanation card with **Back / Next / Skip**.

There are two tours, and the editor tour comes in two lengths:

- **Editor tour** — auto-starts on first launch. The welcome screen offers a
  **Quick start** (short, functional-only) or the **full tour**: canvas, toolbar
  tools (Select / Add Vertex / Add Edge), the Vertices/Edges sidebars, setting a
  phase, and Start Derivation. Quick start drops the educational steps so a
  returning user can skip the explanations.
- **Proof-mode tour** — auto-starts the first time a derivation is started:
  rewrites panel, magic wand, Z/X identity choice, and the proof-steps view.

Both can be replayed any time from **Help → Interactive Tutorial**
(Quick Tour / Full Tutorial), and auto-start on launch can be toggled in
**Preferences → Show tutorial on startup**.

## Design

Per the issue, the feature is kept almost entirely in one new file,
`zxlive/tutorial.py`, so the rest of the codebase only needs a few small hooks.
It's a *guided* tour (read-and-click-Next) rather than wait-for-action, so it can
never leave the app in an inconsistent state: the overlay is modal (it grabs the
keyboard and swallows clicks), and step targets are resolved defensively — a
missing/collapsed widget just falls back to a centred card instead of crashing.

Tours are declarative lists of steps, so the quick/full split is just a flag on
each step (educational steps are marked `full_only` and filtered out of Quick
start), and adding further tours later is additive.

Touches:
- `zxlive/tutorial.py` — overlay, step model, controller, both tours (new)
- `zxlive/mainwindow.py` — Help-menu Interactive Tutorial submenu + proof-tour trigger in `new_deriv`
- `zxlive/app.py` — deferred first-run trigger
- `zxlive/settings.py` + `zxlive/settings_dialog.py` — "Show tutorial on startup" toggle
- `test/test_tutorial.py` — pytest-qt coverage (new)
- `CHANGELOG.md`

## Screenshots / recording

https://github.com/user-attachments/assets/50844f53-e978-4823-8ba5-44c861ab880f

## Testing

- `pytest` — added `test/test_tutorial.py` (navigation, skip/finish, first-run
  one-shot gating, proof auto-start, close-mid-tour, replay-cancels-previous,
  quick/full subset and quick-start switching).
- `mypy zxlive test` — no new errors.
- `ruff` — clean.

## Notes / out of scope

- The tour is intentionally passive (click-Next). Optional "try it yourself"
  interactive steps would require a click-through overlay, which trades away the
  modal robustness above — happy to do that as a follow-up if preferred.
- Per maintainer feedback, an educational track teaching the basic rewrites on
  real examples (e.g. 3 CNOTs = SWAP, phase gadgets, graph states, teleportation)
  is a planned follow-up; I've left a couple of questions on the issue to settle
  the approach first.


harsh-791:
@RazinShaikh this implements #523 — guided overlay, editor + proof tours, replay from Help, all in one tutorial.py with minimal hooks. So could you review this PR and share any feedback or suggested improvements. Tests + mypy + ruff are green.

harsh-791:
@lia-approves , Thanks for the steer! A few questions to make sure I build the right thing:

1. **Surfacing the modes** — would you prefer the quick-start vs educational
   choice as a one-time prompt on first launch, or as separate entries under
   Help (e.g. "Quick tour" / "Learn the basics")?
2. **Educational steps** — are you happier with the tutorial *demonstrating*
   rewrites by applying them step-by-step (robust, click-through), or having the
   user perform each rewrite themselves before advancing?
3. **Examples** — should the example circuits live as bundled files (e.g. an
   `examples/` folder) that could also replace the arbitrary startup diagram?
   And is "3 CNOTs = SWAP" a good first one to land, with the rest following?

test/test_tutorial.py
```diff
@@ -0,0 +1,173 @@
...
+def test_editor_tour_navigation(app: MainWindow) -> None:
+    start_editor_tutorial(app)
+    tut = app._active_tutorial  # type: ignore[attr-defined]
+    assert isinstance(tut, Tutorial)
+    assert tut.overlay is not None
+    assert tut.overlay.isVisible()
+
+    n = len(tut.spec.steps)
+    assert tut.index == 0
+    assert not tut.overlay.back_button.isEnabled()
```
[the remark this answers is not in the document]
96-LB: Copilot is correct—this test fails.

zxlive/tutorial.py
```diff
@@ -0,0 +1,740 @@
...
+        # back to a centred card instead.
+        visible = widget.visibleRegion().boundingRect()
+        if (widget.width() < _MIN_TARGET or widget.height() < _MIN_TARGET
+                or visible.width() < _MIN_TARGET or visible.height() < _MIN_TARGET):
+            return None
+        top_left = widget.mapToGlobal(QPoint(0, 0))
+        local = self.overlay.mapFromGlobal(top_left)
+        rect = QRect(local, widget.size())
+        rect = rect.adjusted(-_SPOTLIGHT_PAD, -_SPOTLIGHT_PAD,
+                             _SPOTLIGHT_PAD, _SPOTLIGHT_PAD)
```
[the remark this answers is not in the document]
96-LB: Feel free to ignore this comment; your current behaviour looks fine to me.

zxlive/mainwindow.py
```diff
@@ -262,8 +262,17 @@ def native_shortcut(key: QKeySequence.StandardKey) -> str:
...
+            "&Quick Tour", lambda: self.start_tutorial(quick=True), None,
+            "A short, functional tour of the interface")
+        full_tutorial_action = self._new_action(
+            "&Full Tutorial", lambda: self.start_tutorial(quick=False), None,
+            "The complete guided onboarding tour")
         help_menu = menu.addMenu("&Help")
         help_menu.addAction(user_guide)
+        tutorial_menu = help_menu.addMenu("&Interactive Tutorial")
+        tutorial_menu.addAction(quick_tour_action)
+        tutorial_menu.addAction(full_tutorial_action)
```
96-LB: Right now, choosing "full tutorial" still brings up the prompt asking which tutorial to follow . Perhaps it should skip that dialog, or you could just merge the two options since the user gets the choice anyway.

Additionally, there's currently no way to start the proof tutorial from the menu.

zxlive/tutorial.py
```diff
@@ -0,0 +1,740 @@
...
+        self._pulse_anim.valueChanged.connect(self._on_pulse)
+
+        self._card = _build_card(self)
+        self.title_label: QLabel = self._card.findChild(QLabel, "tutorial_title")  # type: ignore[assignment]
+        self.body_label: QLabel = self._card.findChild(QLabel, "tutorial_body")  # type: ignore[assignment]
+        self.progress_label: QLabel = self._card.findChild(QLabel, "tutorial_progress")  # type: ignore[assignment]
+        self.quick_button: QPushButton = self._card.findChild(QPushButton, "tutorial_quick")  # type: ignore[assignment]
+        self.skip_button: QPushButton = self._card.findChild(QPushButton, "tutorial_skip")  # type: ignore[assignment]
+        self.back_button: QPushButton = self._card.findChild(QPushButton, "tutorial_back")  # type: ignore[assignment]
+        self.next_button: QPushButton = self._card.findChild(QPushButton, "tutorial_next")  # type: ignore[assignment]
```
96-LB: Rather than using `#type: ignore` here, it would be best to properly handle the case where the child is not found. Right now, if there's a bug in the code, the error will pop up at unexpected points because the variables are `None` but they are not properly type-annotated.

You could either check here and fail early if the required children are not found, or change the type of the attributes to `QPushButton | None` and introduce null-checks whenever you use them.

zxlive/tutorial.py
```diff
@@ -0,0 +1,740 @@
...
+
+        painter.fillPath(path, QColor(0, 0, 0, _DIM_ALPHA))
+
+        if has_spot:
+            assert spot is not None
+            self._paint_beak(painter, spot)
+            highlight = self.palette().highlight().color()
+            highlight.setAlphaF(self._pulse)
+            pen = QPen(highlight)
+            pen.setWidth(2)
```
96-LB: You can avoid using `type: ignore` here by moving `spot is not None` into the conditional or by guaranteeing that the `spot` variable is non-null with something like:

```suggestion
        path.addRect(self.rect())
        spot = self._spotlight
        spot = spot if spot else QRect(0, 0, -1, -1) # Invalid QRect as fallback
        if spot.isValid():
            hole = QPainterPath()
            hole.addRoundedRect(spot, _SPOTLIGHT_RADIUS, _SPOTLIGHT_RADIUS)
            path = path.subtracted(hole)

        painter.fillPath(path, QColor(0, 0, 0, _DIM_ALPHA))

        if spot.isValid():
            self._paint_beak(painter, spot)
            highlight = self.palette().highlight().color()
            highlight.setAlphaF(self._pulse)
            pen = QPen(highlight)
            pen.setWidth(2)
            painter.setPen(pen)
            painter.setBrush(Qt.BrushStyle.NoBrush)
            painter.drawRoundedRect(spot, _SPOTLIGHT_RADIUS, _SPOTLIGHT_RADIUS)
```

zxlive/tutorial.py
```diff
@@ -0,0 +1,740 @@
...
+# A reference to the running tour is kept on the main window so it is not
+# garbage-collected mid-tour and so only one runs at a time.
+
+def _run(main_window: "MainWindow", spec: TutorialSpec) -> None:
+    existing = getattr(main_window, "_active_tutorial", None)
+    if existing is not None:
+        existing.skip()
+    tutorial = Tutorial(main_window, spec)
+    main_window._active_tutorial = tutorial  # type: ignore[attr-defined]
+    tutorial.start()
```
96-LB, line 1041: It would be better to define `MainWindow._active_tutorial` in the class definition so the type-checker can see it instead of relying on `getattr` to hack in the behaviour.

zxlive/tutorial.py
```diff
@@ -0,0 +1,740 @@
...
+def maybe_start_first_run(main_window: "MainWindow") -> None:
+    """Auto-start the editor tour when the startup setting is enabled.
+
+    The setting is flipped off immediately so the tour shows exactly once per
+    time the user opts in (via Preferences). Replaying from the Help menu does
+    not touch it.
+    """
+    if get_settings_value(SHOW_ON_STARTUP, bool, True):
+        set_settings_value(SHOW_ON_STARTUP, False, bool)
+        start_editor_tutorial(main_window)
```
96-LB, line 1069: If the user quits the app during the tour, it won't show again even if they didn't complete it (or start it...).

Besides, would it perhaps be better to leave `SHOW_ON_STARTUP` untouched so that it shows up on every startup? I think this closer matches the behaviour of other applications. I'm not sure how useful this setting is if it only shows it once on the next startup; if the user just wants to see the tour one time, the buttons you added in the help menu are much more convenient than restarting the app.

zxlive/tutorial.py
```diff
@@ -0,0 +1,740 @@
...
+        self.overlay = TutorialOverlay(self, self.main_window)
+        # Follow the main window's size/position while the tour is running.
+        self.main_window.installEventFilter(self)
+        self.overlay.setGeometry(self.main_window.rect())
+        self.overlay.show()
+        self.overlay.raise_()
+        self.overlay.setFocus()
+        # Grab the keyboard so panel shortcuts (v/e/w/s, ...) and Tab can't
+        # drive the dimmed UI behind the overlay while the tour is running.
+        self.overlay.grabKeyboard()
```
96-LB: I'd recommend avoiding `grabKeyboard()` here, as it's SUPER aggressive. It eats up most of my OS inputs, including alt-tab and the Super key.

96-LB (commented):
Thanks for the PR! It is looking very good so far. I've gone through and left some preliminary review comments.

As a general note, you may want to update the styling a little bit to make sure it looks readable on all platforms. The contrast on the "Step X of Y" text is very poor on my machine:
Image

zxlive/mainwindow.py
harsh-791: Good points — addressed both. I merged the Quick/Full menu items into a single "Editor Tour" so the choice only lives on the welcome card (no more redundant prompt). And I added a "Proof Mode Tour" entry, enabled only while a proof tab is active (same as the export-proof actions) so it always has its targets to spotlight.

zxlive/tutorial.py
harsh-791: Good call. I removed the findChild lookups entirely: the card builder now returns a small typed _CardWidgets dataclass, and the overlay stores properly typed QLabel/QPushButton references — so there's no Optional/# type: ignore and a missing child would fail loudly at construction. (I left the two _active_tutorial ignores: promoting it to a typed attribute turns the suppressed attr-defined into union-attr across ~20 test sites for little gain, but happy to do that too if you'd prefer.)

zxlive/tutorial.py
harsh-791: Thanks, applied your pattern — spot now falls back to an invalid QRect() when there's no spotlight, so the branches just check spot.isValid() with no assert/# type: ignore. (The addRect/addRoundedRect calls also wrap in QRectF now, so those arg-type ignores are gone too.)

zxlive/tutorial.py
harsh-791: Done — MainWindow._active_tutorial: Optional["Tutorial"] is now declared on the class (with a TYPE_CHECKING import), so tutorial.py accesses it directly instead of via getattr, and the # type: ignore[attr-defined] casts are gone. I added two small typed helpers in the tests (active_tutorial/overlay_of) to narrow the new Optional cleanly rather than re-introducing ignores.

zxlive/tutorial.py
harsh-791: Agreed on both — fixed. maybe_start_first_run no longer flips the setting off; it just starts the tour when SHOW_ON_STARTUP is enabled and leaves it as-is, so closing the app mid-tour doesn't permanently hide it, and it shows on every startup until disabled in Preferences (matching the typical "show on startup" behaviour). The Help-menu entries cover the "just once" case.

zxlive/tutorial.py
harsh-791: Good point — removed grabKeyboard() entirely. It was an OS-level grab and eating global shortcuts like alt-tab/Super is unacceptable. The overlay still takes focus (so its own Back/Next/arrow keys work) and swallows mouse clicks; I just dropped the keyboard grab rather than replacing it with anything heavier.

harsh-791:
cc @96-LB

lia-approves (commented):
> @lia-approves , Thanks for the steer! A few questions to make sure I build the right thing:
> 
> 1. **Surfacing the modes** — would you prefer the quick-start vs educational
>    choice as a one-time prompt on first launch, or as separate entries under
>    Help (e.g. "Quick tour" / "Learn the basics")?
> 2. **Educational steps** — are you happier with the tutorial *demonstrating*
>    rewrites by applying them step-by-step (robust, click-through), or having the
>    user perform each rewrite themselves before advancing?
> 3. **Examples** — should the example circuits live as bundled files (e.g. an
>    `examples/` folder) that could also replace the arbitrary startup diagram?
>    And is "3 CNOTs = SWAP" a good first one to land, with the rest following?
>  

Thanks for the questions. @96-LB addressed your first question.  I would guess that for 2., user performed rewrites would be preferred due to being more interactive. For 3., yes, 3 CNOTs being a swap would be great to keep things simple for first-time users. One of the trickier aspects is robustness: Implementation-wise, this would be checking that the user does indeed up to only connectivity matters (OCM): draw the correct diagram for the 3 CNOTs circuit, then does indeed end up with the diagram for the swap.  The tutorial should highlight both what steps to do and where to find them in the GUI.

harsh-791:
@lia-approves — implemented the interactive 3 CNOTs → SWAP lesson in this PR: user-performed steps with OCM checks, bundled examples, Help menu + welcome entry. Proof tour is suppressed during the lesson. 127 tests pass. Ready for another look when you have time!

96-LB:
We have decided to accept a different PR for this issue. Thanks for your work, and best of luck with the rest of unitaryHACK!

PR #543, opened 2026-06-12
in edit and rule modes, add "Adjust Scalar" action to manually adjust the global scalar

### LINKED ISSUE #509: Feature request: allow adjustment of scalar in graph
I'd like to be able to create a custom "Decompose Hadamard into Z-spider" rule from the [ZXBook eq. 3.84](https://zxcalc.github.io/book/html/main_htmlch3.html#x6-67007r84):

Image

But I'm unable to do that because (I think) I need to be able to change the scalar.

dlyongemallo:
Fixes #509.


jvdwetering:
Changing the scalar in proof mode would break the semantics of proof mode. I do think it makes sense to have this behaviour in editing mode and when creating a new rewrite rule.

Issue #270, opened 2024-06-29
Edge curve is not preserved by saving, applying rewrites, etc.


RazinShaikh:
The curve distance for the edge is stored in EItem but not in side the PyZX graph. As a result, the curve information is often lost such as when applying a rewrite or saving to a file. One possible fix would be to introduce edge_data dictionary in the PyZX graph class. This is similar to vdata for vertices, which is currently supported.

PR #544, opened 2026-06-12
fix magic wand unfuse from identity vertices

### LINKED ISSUE #319: Magic wand unfuse node from identity does not work

dlyongemallo:
Fixes #319.


jvdwetering:
Just to clarify: slicing over an identity to remove it is intended behaviour. There should be two distinguished cases:
* If you fully slice over a spider so that it goes across the two edges of the spider, this should be interpreted as slicing away the spider, resulting in an identity removal.
* If you stay on one side of the edges, then this should be interpreted as wanting to unfuse an arity 1 spider (like in the picture of #319 )

dlyongemallo:
I just noticed that the code to draw the edges contains a TODO referring to issue #270 which was supposed to have been closed.

PR #545, opened 2026-06-12
use cubic bezier curves for smoother edge rendering

dlyongemallo:


jvdwetering:
Just an FYI that a lot of the functionality and look of ZXLive is based on that of TikZiT which also uses cubic curves. It might be worthwhile to copy as much as possible the exact parameters from TikZiT so that a diagram imported from there continues to look the same in ZXLive.

https://github.com/tikzit/tikzit

dlyongemallo:
The self-loop behaviour before this change does not match TikZit. It has a weight of `0.707` (`1/sqrt(2)`) compared to TikZit's `1.0`. Changing this would resize existing self-loops in saved files. Is that desirable, or should it be left alone?

jvdwetering:
I think that is such a minor issue that we can just take that change and make it match tikzit.

PR #549, opened 2026-07-12
Bump Python version to 3.12

### LINKED ISSUE #547: zxlive failed with python 3.9
## Error Description 

In the current file `pyproject.toml` the python version requirement is `requires-python = ">=3.9"`. 
Building (and running) with python 3.9, however, gives the following error.
```
No solution found when resolving dependencies for split
  │ (markers: python_full_version == '3.11.*'):
  ╰─▶ Because myst-parser>=5.0.0 depends on sphinx>=8 and only the
      following versions of myst-parser are available:
          myst-parser<=5.0.0
          myst-parser==5.1.0
```

## System Information 

My system is MacOS running on M5. 

## Fix 

ZXlive can run successfully if the version requirement is changed to `requires-python = ">=3.12"`

96-LB:
Closes #547.

Updates the minimum Python version for running ZXLive to Python 3.12 (previously Python 3.9). Also updates the corresponding versions in the tooling and GitHub workflows.

There are various comments throughout the code detailing small changes to make once the code was upgraded to 3.10 or 3.11, which I'm happy to either bundle into this PR or open a separate PR for.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
-        python-version: [ 3.11 ]
+        python-version: [ 3.12 ]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,7 +3,7 @@ name: lint
 on:
   push:
     branches: [ master ]
-  pull_request: 
+  pull_request:
     branches: [ master ]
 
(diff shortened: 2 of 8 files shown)


96-LB:
I added a commit which updates the minimum versions of pyzx and mypy since pyzx v0.10.0 has incorrect typing information that causes some mypy errors. We were stuck on v0.10.0 because future versions dropped support for Python 3.9, but we can upgrade that now. The commit also fixes a mypy error which was introduced by the PR.

jvdwetering:
Regarding the minimum version of PyZX, since there are now monthly releases and ZXlive should always use the newest version of PyZX, we should think about having some automated script that for every monthly release of ZXlive also bumps the version of pyzx needed.

PR #550, opened 2026-07-19
Fix failing CI

96-LB:
Fixes the mypy linting errors in the CI caused by type errors and refactors a couple functions to satisfy complexipy. Also preemptively adds a null check in `MainWindow.active_panel` which is causing a failing test in #530 and blocking merge of that PR.

This PR is currently marked as a draft because it relies on #549. It currently contains the commits from that branch in the diff, but once #549 is merged that should be fixed.

### DIFF
--- a/zxlive/animations.py
+++ b/zxlive/animations.py
@@ -153,9 +153,9 @@ def _morph_graph_to_or_from_center(to_center: bool,
     group = QParallelAnimationGroup()
     for v in moves:
         if to_center:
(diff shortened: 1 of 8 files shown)


jvdwetering:
Okay, #549 is merged, so is this ready to go then?

96-LB:
Yes, I just cleaned up the duplicate commits. Should be ready for merge now!

PR #551, opened 2026-08-03
Fix magic wand Hopf rule matching

96-LB:
Fixes two bugs in the magic wand Hopf rule matching: one was introduced by a refactor in #550 which inverted an if statement, and another was a pre-existing bug which caused the magic wand to fail to match X-spiders connected by multiple Hadamard edges. Additionally cleans up the code in the function a little bit to remove an avoidable `#type: ignore`.


96-LB:
I updated the check to avoid matching on two dummy/boundary nodes per Copilot's comment.

RazinShaikh:
Looks good

